PR #17, opened 2017-11-13
typo

eakoning:


steved0x:
@eakoning Thank you for your contribution, I am merging it now and it should be live shortly. Please accept our apologies for this initial delay while we work though the initial backlog of contributions.

PR #20, opened 2017-11-14
Add the QA about System error 53

lsprado:
Add the QA about System error 53


alexhomer1:
Awaiting verification from product team

alexhomer1:
The dev team response is that there are several reasons this error can occur, not just firewall issues. I will update the source docs and add more information related to this instead of merging this PR. Thank you for your input.

alexhomer1:
Source files have been updated with this information

PR #28, opened 2017-11-20
Update restore-data-different-instance.md

CharlesWB:
Correct a typo where a letter was missing in the title "Required Permissions".


steved0x:
@CharlesWB Thank you for your contribution, I am merging it now and it should be live shortly. Please accept our apologies for this initial delay while we work though the initial backlog of contributions.

PR #11, opened 2017-11-07
Typo in the $(Date:yyyyMMdd) token

jantielens:
The $ was forgotten.

PR #39, opened 2017-12-01
Add missing $

MarcDenman:
All of the other examples have a `$` at the start so I assume this one should as well!


MarcDenman:
Just seen #11 so this PR is unnecessary.

PR #42, opened 2017-12-11
Fix a small typo

Bheesma:


steved0x:
@bheesma Thank you for catching this, I am merging this and it should be live soon.

PR #45, opened 2017-12-13
Correct the example for quotes

hajile00:
The image for the quote example is not produced from the input. Fixed the input to match output.


steved0x:
@hajile00 Thank you for your contribution.

#sign-off

PR #52, opened 2017-12-18
Update comparison-git-tfvc.md

DavidDeSloovere:
Add VSTS as server option for TFV.


DavidDeSloovere:
People had questions is this was possible, because the doc was wrong.

steved0x:
@DavidDeSloovere Thanks for catching this, I am merging this in and it should be live soon. Thank you for your contribution!

PR #55, opened 2017-12-18
Typo in both was boht

tegaaa:
Typo correction.


steved0x:
@tegaaa Thank you for your contribution, I am merging it now and it should be live shortly.

PR #60, opened 2017-12-20
Space (the final frontier)

lavahot:
Added a missing space between words.


steved0x:
@lavahot Thank you for your contribution, I am merging it now and it should be live shortly.

PR #62, opened 2017-12-21
Fix grammatical error.

bdlane:


steved0x:
@bdlane Thanks for catching that! Thank you for your contribution, I am merging it now and it should be live shortly.

PR #68, opened 2017-12-30
Reversed author/commit dates in explanation

fghibellini:
The author date is the one that describes the time of first commit (i.e. should always be the oldest one) and the commit date timestamps any further reapplication of the commit (e.g. --amend).


steved0x:
@fghibellini Thank you for your contribution, I am merging it now and it should be live shortly.  Note that this same content is also located at [How dates work in Git](https://docs.microsoft.com/en-us/vsts/git/concepts/git-dates) - I am also going to make your fix there (and then at some point I will consolidate them into a single article). Thanks again!

PR #76, opened 2018-01-05
Correct a typo

gstevenson:


steved0x:
@gstevenson Thanks for catching this, I am merging it and it should be live soon, Thank you!

PR #81, opened 2018-01-07
Update list of supported TFS versions

hkamel:
Removing TFS 2015 from supported versions, "build Machine Image" is only available starting TFS 2017


steved0x:
@hkamel Thank you, merging now!

PR #87, opened 2018-01-10
Update samples.md

vaughanje:
Fixed misspelling of "library" in title


steved0x:
@vaughanje Thank you for catching this, I am merging it now and it should be live shortly.

PR #90, opened 2018-01-11
Update configure-features-after-upgrade.md

ArieHein:
spelling in the note section of step 2 of the wizard


steved0x:
@ArieHein Thank you for catching this, I am merging it now and it should be live shortly.

PR #96, opened 2018-01-13
Update node.md

mmaitre314:
Fix typo


steved0x:
@mmaitre314 Thank you for catching this, I am merging it now and it should be live shortly.

PR #98, opened 2018-01-13
Removed duplicate statement

hkamel:
In Define Tasks --> Remaining Work section


steved0x:
@hkamel MMI Thank you for your contribution, I am merging it now and it should be live shortly.

PR #99, opened 2018-01-14
Update used WIT

hkamel:
Since we are in the context of CMMI process, it's better to mention CMMI WIT instead of Scrum ones.`


steved0x:
@hkamel Thank you for your contribution, I am merging it now and it should be live shortly.

PR #100, opened 2018-01-15
Update index.md

ArieHein:
spelling


steved0x:
@ArieHein Thank you for catching this, I am merging it now and it should be live shortly.

PR #103, opened 2018-01-16
Reflect the right path for search service

hkamel:
replaced TFS 15.0 with the full name (actual path) "Microsoft Team Foundation Server 15.0"


steved0x:
@hkamel Thank you for your contribution, I am merging it now and it should be live shortly.

PR #105, opened 2018-01-17
Update manual-test-permissions.md

divais:


steved0x:
@divais Thank you for your contribution, I am merging it now and it should be live shortly.

PR #107, opened 2018-01-18
Requirement for TFS 2017 Update 2 and 3

sebastiankvarnstrom:
Documentation is not updated with TFS 2017 Update 2 and 3 for SQL server requirement. I don't no if this info is correct but I suppose it is


steved0x:
@sebastiankvarnstrom Thank you for your contribution, I am sending it over to the right folks for review and will get back to you shortly.

steved0x:
@sebastiankvarnstrom After reviewing with the team, we are going to close this PR without merging. They said the point of the lines that reference updates is to indicate that supported versions changed in that update and all subsequent updates. So as it currently stands it shows that TFS 2017 RTM supported SQL Server 2016 RTM, but starting with TFS 2017 Update 1 it requires SQL Server 2016 SP1 or greater. However we are going to see if we can make that distinction more clear in the article. Thank you again for contributing!

PR #109, opened 2018-01-18
Update commits.md

unlysted:
I changed "then" to "the" as was necessary.


steved0x:
@unlysted Thank you for your contribution, I am merging it now and it should be live shortly.

PR #111, opened 2018-01-18
Typo

cartwrightluke:


steved0x:
@cartwrightluke Thank you for your contribution and for catching this, I am merging it now and it should be live shortly.

PR #116, opened 2018-01-22
Remove unsupported characters from task name

ericsciple:


steved0x:
@ericsciple Thank you for your contribution, I am sending it over to the right folks for review and will get back to you shortly.

steved0x:
@ericsciple Sorry for the delay, thank you for your contribution, I am merging it now and it should be live shortly.

PR #120, opened 2018-01-23
Add missing url to list of whitelist urls

ivara:
Noticed {accountname}.vssps.visualstudio.com also is required


steved0x:
@ivara Thank you for your contribution, I am sending it over to the right folks for review and will get back to you shortly.

steved0x:
@ivara I am merging this PR and it should be live soon. Thank you for your contribution!

PR #131, opened 2018-01-26
Clarified offline page creation

laedit:
And fix image for the .order modification when adding a page


steved0x:
@laedit Thank you for your contribution, I am sending it over to the right folks for review and will get back to you shortly.

steved0x:
@laedit Sorry for the delay, thank you for your contribution, I am merging it now and it should be live shortly.

PR #134, opened 2018-01-27
Added an explanation for VSTO account URL rename validation message

dance2die:
Hi, I was switching two VSTO account names and ran into an error and documented the missing piece.

Please let me know if the documentation is readable/accurate or not.
I will make a change accordingly.

The site I am making a change for can be found here
https://docs.microsoft.com/en-us/vsts/accounts/rename-vsts-account
_____
**Explanation for the change**: 
Updated document explaining that a VSTO owner cannot rename the account URL to the name that was used before within an hour.


steved0x:
@dance2die Thank you for this additional information. This is good information for the readers of this article. The screenshot is very good, but in this case I think we could convey the same information at the top, right after the caution statement, like this:

Caution: ....

>[!NOTE]
>An account cannot be renamed within one hour of the most recent rename.

If we add a note like this, we made need to change the Caution segment as well. If you can make this change (replace your screenshot and text with the NOTE above, I wills end it to the author of that article for approval.

Thank you for your contribution!

dance2die:
Thank you @steved0x for your kind guidance.
I've updated the documentation according to your advice and removed my screenshot file.

steved0x:
@dance2die This looks good, I am merging it in now. Thank you for your contribution!

PR #137, opened 2018-01-30
Fix spelling

jawn:


steved0x:
@jawn Thank you for fixing this, I am merging it now and it should be live shortly. Thank you!

PR #153, opened 2018-02-06
Update tools.md

DanSchatz:
Quick typo


steved0x:
@DanSchatz Thanks for catching this, I am merging it now and it should be live shortly. Thank you!

PR #157, opened 2018-02-07
reference correct tfv permissions md

jiangjianliang:


steved0x:
@jiangjianliang Thank you for catching this, I am merging it now and it should be live shortly. Thank you!

PR #166, opened 2018-02-13
Fix syntax errors in WebPageControl samples

iamwillbar:
Remove spaces inside of tags and incorrectly balanced CDATA tags.


steved0x:
@iamwillbar Thank you for your contribution, I am merging it now and it should be live shortly.

PR #174, opened 2018-02-14
Typo in docs/security/export-users-audit-log.md

agriffard:
Remove duplicated 'the'.


steved0x:
@agriffard Thank you, I am now wondering if we had some automated text replacement to result in these, thank you for helping us fix them :)

PR #180, opened 2018-02-16
Create space between glued together words

jawn:


steved0x:
@jawn Thank you, I am merging this now and it should be live soon.

PR #206, opened 2018-02-25
Replace Logo with Icon

StevenTCramer:
Replace all references to logo with icon.  One could choose logo or icon but given the schema uses icon we should just stick with that wording.


StevenTCramer:
Fix for issue #[202]
Now that I realize I can fix vs just complain. :)

steved0x:
@StevenTCramer Thank you for your contribution, I have assigned this PR to the article author. Thank you.

Issue #221, opened 2018-02-27
Screenshots for Agentless Phase show two tasks in one Agentless Phase


hintbw:
The screenshots for the Agentless Phase show two tasks in one Agentless Phase, but that is not possible. When a user tries to save it will not let the release definition be saved with multiple tasks in one Agentless phase. The screenshots should be updated and it would be useful to call out the current limitation of allowing one one task per agentless phase in text right below the section header.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 54bf6c2f-95c1-943d-26ae-d7099143bccf
* Version Independent ID: 0833bb55-1e54-6953-ee69-de6d191d2ec1
* [Content](https://docs.microsoft.com/en-us/vsts/build-release/concepts/process/phases#agentless-phase)
* [Content Source](https://github.com/MicrosoftDocs/vsts-docs/blob/live/docs/build-release/concepts/process/phases.md)
* Service: unspecified

alexhomer1:
Thank you. I will update the screenshots and check with the development team whether the limitation is just temporary or permanent.

PR #222, opened 2018-02-27
Addition to help in the resolution of my feedback on the docs

hintbw:
While my addition doesn't completely resolve the issue that needs to be fixed - it at least add some additional detail that I believe is important (since the documentation leads one to believe that functionality exists for multiple tasks per agentless phase when that isn't in fact possible.

It is related to the following issue in this repo - https://github.com/MicrosoftDocs/vsts-docs/issues/221 #221


steved0x:
@hintbw Thank you for your contribution, I have assigned this PR to the author of the article for their review. Thank you.

alexhomer1:
Waiting for confirmation from development team as to whether this is just a temporary restriction.

alexhomer1:
It is hoped to remove the restriction soon, so we will update the docs accordingly. Thank you for your input.

PR #233, opened 2018-03-01
Minor text change in pull request comment section

sweenist:
The content in the pre tag mentions pressing the "Enter key."
The result had "return key" without a period.


steved0x:
@sweenist Thank you for your contribution, I am merging this and it should be live soon. Thank you!

PR #242, opened 2018-03-02
Remove spurious &nbsp;

citelao:


steved0x:
@citelao Thank you for fixing this, it looks like an artifact from our migration to docs.microsoft.com. I am merging it now and it should be live shortly. Thank you!

PR #255, opened 2018-03-05
 Updated that it works in VSTS too

ChristopherMank:
In my testing, this does work just fine in VSTS. I was able to download and upload mappings to VSTS.


steved0x:
@ChristopherMank Thank you for checking this out, I am assigning this PR to the author of that article for review, thank you for your contribution.

PR #256, opened 2018-03-05
Updated that field mapping does work in VSTS

ChristopherMank:
I have tested this against my VSTS account and it worked just fine.


steved0x:
@ChristopherMank Thank you for checking this out, I am assigning this PR to the author of that article for review, thank you for your contribution.

PR #260, opened 2018-03-05
fix typo

ckimes89:


steved0x:
@ckimes89 Thanks for fixing this, I am merging it in now and it should be live soon. Thank you!

PR #264, opened 2018-03-05
duplicate "with" in the line

raghul-ravi:
duplicate "with" in the line


steved0x:
@raghul-ravi Thank you for fixing this, I am merging it now and it should be live soon. Thank you!

PR #298, opened 2018-03-09
Update image to show current copy on gate flyout.

Izzmo:
Fixed the image to show the current copy and layout of the gates flyout.


alexhomer1:
Thank you for your input

PR #312, opened 2018-03-12
How to add a member to the Team Project Collection Service Accounts group

MarkIannucci:
This information is available on [serverfault](https://serverfault.com/questions/174108/add-a-user-to-the-team-foundation-service-accounts-group-in-tfs).  It should also be in Microsoft's documentation.

### DIFF
--- a/docs/security/permissions.md
+++ b/docs/security/permissions.md
@@ -128,8 +128,8 @@ You can't remove or delete the default server level groups.
 			</p>
 			<p>
 				If you need to add an account to this group after you install TFS, you can do so using
-				the TFSSecurity.exe utility in the <strong>Tools</strong> subfolder of your TFS installation directory.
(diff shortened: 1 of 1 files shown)


steved0x:
@MarkIannucci Thank you for your contribution, I have assigned this PR over to the author of the article for review. Thank you.

docs/security/permissions.md
```diff
@@ -126,6 +126,11 @@ You can't remove or delete the default server level groups.
 				and not user accounts or groupsthat contain user accounts.
 				By default, this group is a member of <strong>Team Foundation Administrators</strong>.
 			</p>
+			<p>
+				If you need to add an account to this group after you install TFS, you can do so using
+				the TFSSecurity.exe utility in the <strong>Tools</strong> subfolder of your TFS installation directory.
```
MarkIannucci: I don't like the `<strong>` tags, but it seems that the \` for inline code doesn't work in this context (or at least github isn't rendering it)... If the \` works for inline code rendering, please change on merge.

erickson-doug: thanks! ` does indeed work when rendered without our content pipeline. i'll update the merge, and thank you for your contribution!

PR #317, opened 2018-03-13
add note for not supported task group feature

BenjaminAbt:
Right now there is not hint that Task Groups are not supported right now. Just the YAML views shows, that Task Groups are not supprted yet.
image


steved0x:
@BenjaminAbt Thank you for your contribution, I have assigned this PR to the author of the article for review. Thank you.

andyjlewis:
Thanks for the suggestion. I've added a line to the capability table. The change should show up in next day or so.

PR #387, opened 2018-03-22
added bullets to the regions

ManuelGalindo:
Formatting added bullets to the regions


steved0x:
@ManuelGalindo Thank you for updating this, I am merging it and it should be live soon. Thank you!

PR #419, opened 2018-03-27
Added verbiage that rules are system-level

ChristopherMank:
Added verbiage that these rules are not simply form-level, but system-level. Also added that changes to fields do not get added to the history.


steved0x:
@ChristopherMank Thank you for your contribution, I have assigned this PR over to the article author for review. Thank you.

PR #437, opened 2018-03-29
Added Jenkins\VSTS CI\CD AKS deployment tutorial

nicolehaugen79:
I have been collaborating with Micheal Learned to add this content that shows how to setup a Jenkins CI and VSTS CD pipeline that deploys to AKS.  A few things to note:

1.) I used [this existing tutorial](https://docs.microsoft.com/en-us/vsts/build-release/actions/integrate-jenkins-vsts-cicd?view=vsts) that we recently wrote, as a starting point.  As a result:
* I kept in the info on how to configure automatic triggers for building\releasing - not sure if this is too redundant of the info already provided by the previously linked tutorial?
* Since I was told that images should be kept to a minimum, I kept the same images that were in the tutorial that I started from and haven't added any.  Let me know where you think additional images would be beneficial.

2.) This scenario is based on a customer engagement that we did to build a Jenkins + VSTS CI\CD pipeline that deploys to K8s.  One of the questions that came up in this engagement was how to update the image tags rather than using "latest". Our scenario in the engagement was more complex than this - I simplified the scenario to show how you can tokenize the image tag and use Jenkins vars from within VSTS.  There are many ways to do this, but this was essentially the approach that we took in the customer engagement.

Let me know any feedback that you may have.

### DIFF
--- a/docs/build-release/actions/jenkins-vsts-aks-release.md
+++ b/docs/build-release/actions/jenkins-vsts-aks-release.md
@@ -1,0 +1,445 @@
+---
+title: Integrate your Jenkins CI builds with VSTS DevOps | Microsoft Docs
+description: Set up continuous integration (CI) and continuous deployment (CD) for your apps using Jenkins and Visual Studio Team Services (VSTS)
+author: nicolela
+manager: ruisun
+editor: 
+tags: 
+
+ms.assetid: 
+ms.service: 
+ms.devlang: na
+ms.topic: tutorial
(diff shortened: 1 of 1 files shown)


steved0x:
@nicolehaugen79 Thank you for your contribution, I'll assign this to @mlearned and he can merge it and add it to the TOC.

mlearned:
I moved this to our private repo, made modifications, and published the tutorial, its live.

PR #444, opened 2018-03-30
Updated verbiage on update

ChristopherMank:
Existing doc said when work item is updated, but it's only when work item is commented on.


steved0x:
@ChristopherMank Thank you for that clarification, I am merging it in now and it should be live soon. Thank you.

PR #462, opened 2018-04-02
Updated notes on how deletes work

ChristopherMank:
Updated Note to indicate that removing custom field from form does NOT delete the data. The warning box when removing a custom field from the form also indicates that data is NOT deleted.

Also, removed verbiage that fields must be removed from forms before they can be deleted. Deleting fields removes them from forms automatically.


steved0x:
@ChristopherMank Thank you for the update, I have assigned the PR to the article author for review. Thank you.

PR #472, opened 2018-04-04
Add Secure File download location information

alexkokkinos:
Adds information about where the secure file is downloaded.

### DIFF
--- a/docs/build-release/tasks/utility/download-secure-file.md
+++ b/docs/build-release/tasks/utility/download-secure-file.md
@@ -20,7 +20,7 @@ monikerRange: 'vsts'
 
 Use this task to download a [secure file](../../concepts/library/secure-files.md) from the server during a build or release.
 
-Once downloaded, the secure file is located in the `$env:TEMP` directory.
+Once downloaded, the secure file is located in the `$env:TEMP` directory of the VSTS Agent.
 
 ## Arguments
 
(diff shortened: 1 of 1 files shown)


docs/build-release/tasks/utility/download-secure-file.md
```diff
@@ -20,6 +20,8 @@ monikerRange: 'vsts'
 
 Use this task to download a [secure file](../../concepts/library/secure-files.md) from the server during a build or release.
```
madhurig: Can you also add a note that the full path to the downloaded file is accessible via $env:DOWNLOADSECUREFILE_SECUREFILEPATH. If multiple versions of the task are used the variables are disambiguated as DOWNLOADSECUREFILE1_SECUREFILEPATH, DOWNLOADSECUREFILE2_SECUREFILEPATH ...

alexkokkinos: Sure, I can do that!

steved0x:
@alexkokkinos @madhurig Thank you for your contribution and review, I am merging this now and it should be live soon. Thank you!

PatMyron:
Thanks for adding this! The powershell syntax was a bit confusing as someone unfamiliar with powershell. Could we make the syntax less powershell specific?

PR #473, opened 2018-04-04
Add info about the Download Secure File task

alexkokkinos:
This page seems to act as a general overview of Secure Files, but does not provide any information about how to consume the secure files outside of integrating the functionality in a custom task. I've included a link to the Download Secure File Utility task in order to provide this information.


steved0x:
@alexkokkinos Thank you for your contribution, I am merging this now and it should be live soon. Thank you.

PR #488, opened 2018-04-05
Update wrong "Path to publish"

compuguru:
I think that the `Source folder` should be `$(Build.SourcesDirectory)` instead of `$(Build.ArtifactStagingDirectory)`. Otherwise, you're copying files from the same place you're dropping them.

This is for the "C++ app" section in the "Artifacts in Team Build"


steved0x:
@compuguru Thank you for your contribution, I have assigned this to the author of the article for review. Thank you.

PR #489, opened 2018-04-05
Fix "Add New Item" instructions

r0bfr:
Having the user attempt to add a User Control as a way of opening the "Add New Item" dialog is confusing.  Especially when the "User Control" item they need to click is right next to the "New Item" item anyway,


steved0x:
@r0bfr That does look a little simpler and takes you to the same place. I'm merging this and it should be live soon. Thank you for your contribution.

PR #499, opened 2018-04-07
Fix backslash typo

roscoe:
Replaces forward slash with backslash and conforms the code span to the usual style.


steved0x:
@roscoe Thanks for the update, I am merging it now and it should be live soon. Thank you.

PR #511, opened 2018-04-09
fix: upstream-sources link from absolute to rel

nschonni:
Was resolving to 
https://docs.microsoft.com/en-us/concepts/upstream-sources.md 
instead of
https://docs.microsoft.com/en-us/vsts/package/concepts/upstream-sources


steved0x:
@nschonni Thank you for updating that, I'm merging this and it will be live soon. Thank you.

PR #551, opened 2018-04-14
Remove dup entry for Chart for Test plans on Widget Catalog

PeterAntal:


steved0x:
@PeterAntal Thanks for fixing that, I am merging this and it should be live soon.

PR #554, opened 2018-04-16
Normalize Line endings

nschonni:
Since I ran into this elsewhere when switching machines, I just followed https://stackoverflow.com/a/15646791/455535 so that line endings are normalized on checkin


nschonni:
Rebased

nschonni:
No problem, they should have both started as issues. Just wasn't sleeping, and this type of PR is my counting sheep

PR #568, opened 2018-04-17
(ForumsTeam) - Removed Azure Classic Portal Link

v-vijisa:
I have a removed the Azure Classic Portal link, which is deprecated.


steved0x:
@v-vijisa Thank you for your contribution, I have merged this and it should be live soon. Thank you!

PR #604, opened 2018-04-24
Incorrect spacing in script entry

eakoning:
When copying the code like that running npm start refreshVSToken does not work.


steved0x:
@eakoning Thanks for catching that, I am merging it now and it should be live soon.

PR #659, opened 2018-04-29
Rename taskIcon.png to icon.png

niofire:
The icon does not get picked up when naming it as taskIcon.png. Renaming it to limit confusion.


steved0x:
@niofire Thank you for your contribution! I have assigned this to the article author for review.

elbatk:
I made this exact change today before seeing this - so no worries! It should be live within the day, thanks for the heads up @niofire

PR #677, opened 2018-05-02
Removed rogue 'can'

billpratt:


steved0x:
@billpratt Thanks for catching this, I am merging it right now and it should be live soon, thank you.

PR #681, opened 2018-05-02
Update index.md

DarqueWarrior:
Fixing broken Resource link at the bottom for Work Item customization.


steved0x:
@DarqueWarrior Thanks for catching that, I'm merging it now and it should be live soon.

PR #683, opened 2018-05-02
Correcting a copy and paste error.

DarqueWarrior:
Correcting a copy and paste error. Both What is Scrum and What is Agile Development point to what is agile development. Clearly the What is scrum should point to what is scrum doc.


steved0x:
@DarqueWarrior  Thanks for catching that, I'm merging it now and it should be live soon.

PR #690, opened 2018-05-03
chore: Bad trailing underscore in TFVC to Git link

nschonni:


steved0x:
@nschonni Thanks for catching that, I'm merging it now and it should be live soon.

PR #695, opened 2018-05-03
fix: `pre` tag casing in Markdown Guidance

nschonni:


steved0x:
@nschonni  Thanks for catching that, I am merging it now and it should be live shortly.

PR #696, opened 2018-05-03
fix: Bad absolute link to WIQL Reference docs

nschonni:
Not sure if they should both be pointing to the same resource, but I went by one pointing to an absolute URL and the other kept the `.md` extension.


steved0x:
@nschonni Thanks for your contribution, I am assigning this to the article author for review. Thank you.

PR #700, opened 2018-05-03
Fix typo

olsh:


steved0x:
@olsh  Thanks for catching that, I am merging it now and it should be live shortly.

PR #715, opened 2018-05-04
clarify availability of linking a var group to an env

TheBeardedLlama:


TheBeardedLlama:
always a pleasure @alexhomer1 !

PR #718, opened 2018-05-04
Fix typo in H1 heading

wpschaub:
Minor typo fix in the H1 heading of the article. Change Exoplore to Explore.


steved0x:
@wpschaub Thanks for catching that, I am merging it now and it should be live shortly.

PR #719, opened 2018-05-04
Minor tweak to Excel limitations

ChristopherMank:
Previously stated you couldn't import/export test cases, which isn't true. You can do this on test cases. What you can't do, however, is import/export test case steps.


steved0x:
@ChristopherMank Thank you for the contribution, I am assigning this to the article author for review.

PR #761, opened 2018-05-10
updated deprecated method

lisawong19:
The old method name for oAuth doesn't work anymore. Updated it with new method name.


lisawong19:
@elbatk: can you approve of the change?

elbatk:
Thanks @lisawong19  :)

PR #763, opened 2018-05-11
Update link for Code Search

JimsonChalissery:
Update the h/w requirements link for Code Search to point to the TFS 2017 documentation of Code Search rather than the VSTS version.

### DIFF
--- a/docs/tfs-server/requirements.md
+++ b/docs/tfs-server/requirements.md
@@ -431,7 +431,7 @@ agents on your TFS application tiers.
 
 If you plan to use Code Search, we typically recommend setting up a separate server for it. 
 For more details, see 
-[hardware requirements for Code Search](../search/code/administration.md#hardware-recommendations).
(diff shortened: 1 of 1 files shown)


docs/tfs-server/requirements.md
```diff
@@ -431,7 +431,7 @@ agents on your TFS application tiers.
 
 If you plan to use Code Search, we typically recommend setting up a separate server for it. 
 For more details, see 
-[hardware requirements for Code Search](../search/code/administration.md#hardware-recommendations).
+[hardware requirements for Code Search](../search/code/administration?view=tfs-2017).
```
tapasrs: We should have this view as tfs-2018

JimsonChalissery: This particular link is from VSTS version of TFS requirements, which by itself is strange. Ideally each version of TFS doc should point to its corresponding Code Search doc.

nschonni:
Looks like this file moved to a separate/private repo

PR #765, opened 2018-05-11
changed source folder

raaaimund:
Should not be the source folder of C++ app example $(Build.SourcesDirectory) instead of $(Build.ArtifactStagingDirectory).


andyjlewis:
Yeah, looks like a typo. Other writers hold the token on this topic so adding them rather than merging right now to avoid churn.

andyjlewis:
Just realized this change is to artifacts topic and I don't believe the runway is clear, so merging...

andyjlewis:
Oh, and Raimund, thank you for the fix!

PR #773, opened 2018-05-11
Fix index and align with toc for articles

wpschaub:
- Remove black line in index
- Align title to use expose, not exposure, consistently


steved0x:
@wpschaub Thank you for fixing these, I am merging this PR and it should be live soon.

PR #802, opened 2018-05-16
Highlight an additional limitation when using Git LFS with VSTS.  If …

MarkIannucci:
…this were in this guide we may have decided not to use it.

https://developercommunity.visualstudio.com/content/problem/252830/dragging-and-dropping-a-binary-file-configured-for.html


steved0x:
@MarkIannucci Thank you for your contribution, I am merging this now and it should be live soon. Thank you.

PR #831, opened 2018-05-22
Update transforms-variable-substitution.md

jconrey:
removed redundant wording


alexhomer1:
Thank you for your feedback. While we are migrating the documentation to a new site, we are unable to accept direct PRs. However, I have applied the changes to the source topics and they will be published when the migration is complete.

PR #839, opened 2018-05-24
fix: unclosed INCLUDE

nschonni:
Maybe have been intentional since it's sitting behind a custom TODO block


steved0x:
@nschonni We are moving some documents around internally and the build-release folder is one of them, so I'm going to close this PR but I am forwarding it to the author so they can fix it in the new location. My guess is that since it was commented out they didn't look at the syntax (possibly a cut and paste?).

PR #852, opened 2018-05-24
fix: rest-api-overview pointing at vs.com/api

nschonni:
Maybe a find/replace accident when the files where moved to `previous-api` folder


steved0x:

PR #863, opened 2018-05-24
Add uninstall 2018 information for Elasticsearch service

chad-ramos:
Path and bat file name have changed from 2017 to 2018.  Add explicit information for 2018.


alexhomer1:
Thank you for your feedback. I will pass this on to the product team and update the topic based on their response.

PR #868, opened 2018-05-25
fix: Link to kerberos constrained delegation

nschonni:
Seems like a copy/paste of a partial link thing that I do sometimes too.

/cc @KathrynEE


steved0x:

PR #873, opened 2018-05-25
fix: absolute to relative link for link

nschonni:
Previous liking didn't work because of the absolute link with the `.md` extension in the link. Alternate fix is just to drop that and keep the absolute link


steved0x:

PR #919, opened 2018-05-30
fixed "OS that supported by VS 2013" link

shayki5:
The link was inccorect and redirect to: https://www.visualstudio.com/vs/.
I fixed it to the correct link.


shayki5:
Ok, thanks :)

PR #929, opened 2018-05-31
fix: VsTest folder renamed to V2 on GH

nschonni:
The folder was renamed in the GH project so they were all 404ing.
There was a case where there was an explicit reference to the V1, but it used a specific release branch, so it was still working.


steved0x:
@nschonni Thank you for your contribution, I am merging it now and it should be live soon.

PR #932, opened 2018-05-31
fix: AzureRmWebAppDeployment renamed to V4 on GH

nschonni:
Since this is in the "archive" folder maybe you want to point to an older version, but this fixes it for now.


steved0x:
@nschonni Thank you for your contribution, I'm assigning this to the article author to see if there is a better link to an older version or if this new version s the right one.

PR #935, opened 2018-05-31
fix: Drop F12 from devtools-guide links

nschonni:
Probably got renamed at some point, not sure if it's better to fix the upstream docs repo to put in a redirect though


nschonni:
Think the redirect was broken by https://github.com/MicrosoftDocs/edge-developer/commit/2a4f23fb61af540ffbf1660956089eca24ac60b6

KathrynEE:
@nschonni - Thank you for the correction.

PR #936, opened 2018-05-31
fix: Blockquote wrapping split link ending

nschonni:


steved0x:

PR #937, opened 2018-05-31
fix: Period instead of slash before Collection

nschonni:
Wrapped in backticks like other non-functional links


steved0x:

PR #938, opened 2018-05-31
fix: Remove bad title text for Code Churn image

nschonni:


steved0x:

PR #939, opened 2018-05-31
fix: RunDistributedTests renamed to V1 on GH

nschonni:


steved0x:

PR #940, opened 2018-05-31
fix: SqlAzureDacpacDeployment renamed to V1 on GH

nschonni:
Think some stuff got lost in the folder rename and rebase from your private repo, so this partially re-does #841


steved0x:

PR #957, opened 2018-06-01
Add release date and version of TFS 2018.2

pascalberger:
Add release date and version of TFS 2018.2


steved0x:
@pascalberger Thanks for updating that, I am merging this PR now and it should be live soon. Thank you!

PR #985, opened 2018-06-05
Updated Authentication section

ChristopherMank:
Added detail to authentication section to indicate that permission from PAT is only used at time of register and to actually configure the agent on the server, you must be a local admin.

### DIFF
--- a/docs/pipelines/agents/agents.md
+++ b/docs/pipelines/agents/agents.md
@@ -163,7 +163,7 @@ as shown in the following schematic.
 
 ## Authentication
 
-To register an agent, you need to be a member of the [administrator role](pools-queues.md#security) in the agent pool. Your agent can authenticate to VSTS or TFS using one of the following methods:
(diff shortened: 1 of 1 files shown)


docs/pipelines/agents/agents.md
```diff
@@ -163,7 +163,7 @@ as shown in the following schematic.
 
 ## Authentication
 
-To register an agent, you need to be a member of the [administrator role](pools-queues.md#security) in the agent pool. Your agent can authenticate to VSTS or TFS using one of the following methods:
+To register an agent, you need to be a member of the [administrator role](pools-queues.md#security) in the agent pool. Your agent can authenticate to VSTS or TFS using one of the following methods. The identity of agent pool administrator is needed only at the time of registration and is not persisted on the agent, nor is used in any further communication between the agent and VSTS or TFS. Also, one needs to be a local administrator on the server in order to configure the agent.
```
vijayma: @ChristopherMank The same line is present in the "Communications" section already. I guess you are suggesting that we add the same line to the "Authentication" section as well. If so, can you move this line to before "Your agent can authenticate to VSTS or TFS using one of the following methods."

ChristopherMank:
I updated the location of the wording as you suggested @vijayma

PR #1008, opened 2018-06-08
Changed name to something more appropiate

theterminator3000:


steved0x:
@theterminator3000 Willy Schraub is a contributor to this article and I think that's why his name is used for the example. Rather than modify his name, I changed the example to use one of our approved fictional names: Jamal Hartnett (jamal@fabrikam-fiber.com). I'm merging the PR and it should be live soon. Thank you for your contribution.

PR #1010, opened 2018-06-08
typo: double "security"

nschonni:


steved0x:
@nschonni I need to check with the vendor that is merging these back to our private repo - I remember this one - the double security was an error, but the fix was to update the path to "/vsts/organizations/security/get-started-stakeholder" because that file had moved. I made that change (well I need to run it down and see where I made it so I can figure out what happened), but now it is back to the "old" way and I need to figure out what happened.

PR #1014, opened 2018-06-09
Fixing the link to copy files repository 

yashints:
The copy files repository has been moved but the link is pointing to the old repo


steved0x:
@yashints Thank you for making this update, I am merging it now and it should be live soon. Thank you.

PR #1070, opened 2018-06-17
fix: dead links in Release Notes

nschonni:
Just some things that aren't redirecting anymore since there has been a few relocations over time.


steved0x:
Thanks for catching these!

PR #1071, opened 2018-06-17
Fix img tag for creating a new service bus.

jimmylewis:
The tag had an invalid character in the markup, which caused the markup to render as text in the document.


steved0x:
@jimmylewis Thank you for your contribution, I am merging this now and it should be live soon.

PR #1073, opened 2018-06-17
fix: Double "security"

nschonni:
Redo for #1010 which appears to have disappeared


steved0x:
@nschonni I thought I fixed this but it is still there, merging now and keeping an eye on it to make sure it stays fixed.

PR #1094, opened 2018-06-20
typo: Missing "ref" in tfs/server redirects

nschonni:
Not sure if it's better to do a redirect on the `tfs-docs` repo, but that appears to be private right now


steved0x:
@nschonni It looks like the redirects are all here, so this is the right place. I am merging these now.

PR #1105, opened 2018-06-20
Typo 'experience'

michha:


steved0x:
@michha Thanks for catching that, I'm merging it now and it should be live soon.

michha:
Live Link: https://docs.microsoft.com/en-us/vsts/release-notes/2018/jun-19-vsts#build-and-release-1
so I can check when its live

PR #1151, opened 2018-06-26
Fix release notes links

nschonni:
Maybe the redirects got shut down temporarily but they weren't working for me so I just did some find and replace in the release notes:
- visualstudio.microsoft.com/articles/news -> relative links
- https://visualstudio.microsoft.com/docs and /en-us/docs to /vsts/ links


steved0x:
@nschonni Thanks for fixing these, I'm merging this now.

PR #1165, opened 2018-06-27
chore: Add language to code manifest code fences

nschonni:
Minor, but was set on some and not others


steved0x:
@nschonni Nice, I have sent this to the author so they can merge it, since I'm not as familiar with this area. Looks like a nice catch though.

elbatk:
Thanks @nschonni :)

PR #1168, opened 2018-06-27
Added a note on the deleted Key scenario

ManuelGalindo:
A customer asked this question and why it is not on the documentation


alexhomer1:
Thank you for your feedback. I have passed this on to the product team for confirmation.

PR #1179, opened 2018-06-28
Update options.md

sevensc:
.rr is not working, use rr instead


fakhrulhilal:
`.r` is working for me. It means, it will prepend the dot character before incremented number. Current documentation is already correct.

PR #1195, opened 2018-07-02
correction on text

vitorrossetto2:


steved0x:
@vitorrossetto2 Thanks for fixing this, I am merging it and it should be live soon.

PR #1220, opened 2018-07-05
fix: Add language to codefence

nschonni:
- mostly adding `xml` and fixing some minor formatting that the syntax highlighting showed
- ps replaced with powershell for better VS Code highlighting

Still seeing some big diffs from file encoding, but if you switch to the "rich" diff it shows the actual change

### DIFF
--- a/docs/pipelines/agents/proxy.md
+++ b/docs/pipelines/agents/proxy.md
@@ -103,7 +103,7 @@ In the agent root directory, create a .proxy file with your proxy server url.
 
 # [Windows](#tab/windows)
 
-```ps
+```powershell
 echo http://name-of-your-proxy-server:8888 | Out-File .proxy
 ```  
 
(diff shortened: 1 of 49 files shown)


steved0x:
@nschonni What's the benefit of using powershell instead of ps for the code fences for those samples? Something to do with VS Code? Can you put a screenshot of the differences here if you can, before I pass this PR along. Thank you!

nschonni:
@steved0x Looks like GitHub also doesn't recognize `ps` as a valid language ID either. See https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/agents/proxy.md#tab/windows vs the "Rich Diff" in the PR

nschonni:
@steved0x did you need any other info?

steved0x:
@nschonni I'm merging this, thank you!

PR #1236, opened 2018-07-06
Update bulk-add-modify-work-items-excel.md

dend:
Update the link to a working plugin.


steved0x:
@dend Thank you for giving a better link, I am merging this now and it should be live in a few days.

PR #1244, opened 2018-07-07
Update scale-teams.md

hkamel:
Removing extra link and fix article link was redirected to 404. 

- For now, I have added the full URL as I couldn't identify the source document in GitHub


steved0x:
@hkamel Thanks for removing that duplicate link, I restored the other link back to its original form since it is working correctly and I am merging this PR. Please let me know if that link still doesn't work for you (404)

Issue #1273, opened 2018-07-11
New REST APIs Link don't work on Microsoft doc.


Powerhelmsman:
The issue was created by @sasonarum from https://github.com/MicrosoftDocs/feedback/issues/511
Describe the bug
On Microsoft docs at: https://docs.microsoft.com/en-us/vsts/integrate/concepts/dotnet-client-libraries?view=vsts
The "New REST APIs" link isn't working it gives 404
To Reproduce
Go to 'https://docs.microsoft.com/en-us/vsts/integrate/concepts/dotnet-client-libraries?view=vsts'
Click on 'New REST APIS' link
Error: PAge not found (404)
Expected behavior
Screenshots

Desktop (please complete the following information):
OS: Windows 10
Browser: Chrome
Version: Version 67.0.3396.99 (Official Build) (64-bit)
Additional context
We are upgrading from TFS 2012 to TFS 2018 and we used teh REST APIs to connect to XAMLE build definitions now we need to connect to JSON builds from our .net program and I need all the REST APIs reference for that. Additional information is appreciated.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 413163f0-28c3-a2a7-d626-fcb81bcb2257
* Version Independent ID: 57c127da-2b1e-d29e-9abb-a3c258762ccf
* Content: [.NET Client Libraries for VSTS (and TFS)](https://docs.microsoft.com/en-us/vsts/integrate/concepts/dotnet-client-libraries?view=vsts)
* Content Source: [docs/integrate/concepts/dotnet-client-libraries.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/integrate/concepts/dotnet-client-libraries.md)
* Product: **devops**
* GitHub Login: @elbatk
* Microsoft Alias: **elbatk**

eriawan:
@Powerhelmsman 

I'm updating this doc page now, and I'll submit a PR soon.

Yes, the link to TFS/VSTS REST API has changed.

eriawan:
@Powerhelmsman 

A fix this issue has been submitted on PR #1283. 

@steved0x could I search for other incorrect links of TFS/VSTS REST API and update them in one PR?

Many other VSTS doc pages are still having the incorrect link.

eriawan:
@steved0x 

I think this issue now can be closed. 🙂

elbatk:
[...]

PR #1283, opened 2018-07-12
Solve #1273 to correct link of TFS/VSTS REST API

eriawan:
Update dotnet-client-libraries.md to have correct link of TFS/VSTS REST API.
This solves issue #1273.

cc @elbatk and @steved0x


steved0x:
@eriawan Thanks for fixing this, I have merged it and it should be live shortly.

PR #1308, opened 2018-07-16
typo: Double word "appear"

nschonni:


steved0x:
@nschonni Thanks for catching these, I thought we had them already, I am going to note these and make sure they don't revert.

PR #1330, opened 2018-07-18
Fix Artifacts Documentation Formatting

jpreese:
It looks like this got malformed in commit c4fe4604cc49cd3f0864f125898c956ac350f8ba

### DIFF
--- a/docs/pipelines/release/artifacts.md
+++ b/docs/pipelines/release/artifacts.md
@@ -14,7 +14,7 @@ monikerRange: '>= tfs-2015'
 
 # Release artifacts and artifact sources
 
-[!INCLUDE [ver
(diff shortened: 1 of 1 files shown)


alexhomer1:
Thank you for your input!

jpreese:
@alexhomer1 this may seem like a silly question, but whats the release cadence for this repository? It's been close to two weeks and I still haven't seen this change live. Thanks!

alexhomer1:
It was merged the same day, but probably conflicted with an edit to the internal repo and so the change was lost. This does happen very occasionally and typically is fixed manually afterwards by the administrator. Anyway, I have made the edit to the source doc today - the change should be live within 24 hours. Thank you for your help!

PR #1338, opened 2018-07-18
Moved "argument" outside of inline code delimiters

draevin:
Improving clarity; the argument is "/target:clean" not "/target:clean argument" and should be represented as such to avoid confusion and maintain consistent formatting.


steved0x:
@draevin Thanks for fixing that, I am merging this now and it should be live soon.

PR #1346, opened 2018-07-19
Update links to docs.microsoft.com

flcdrg:
Original links were no longer working


steved0x:
@flcdrg Thanks for fixing that, I am merging it now and it should be live soon

PR #1353, opened 2018-07-20
Update add-external-user.md

hkamel:
Fixing markdown issue, currently displayed as below 

image


steved0x:
@hkamel I'm going to close this PR and assign a work item to that author to fix internally, that way we can stage it and make sure it is fixed before we merge. Thanks for letting us know.

PR #1356, opened 2018-07-20
docs: Reformat for bullet list rendering

nschonni:
The bullet list doesn't support the multi-line text of the previous version. Pulled text from linked articles.

___
I'm not a tech writer so maybe someone has a better idea of wording for this.


elbatk:
this is fine, @nschonni ! These docs are for older APIs and don't get as much love so thanks for the fix.

nschonni:
@elbatk looks like this got reverted in the latest be shuffle

elbatk:
@nschonni working in a branch today that should readd them, thanks!

PR #1368, opened 2018-07-20
chore: Remove empty EXE

nschonni:
Empty file, and probably good to exclude the `exe` stuff from the repo


steved0x:
@nschonni Yes thank you.

PR #1370, opened 2018-07-20
Update for new location of OAuth Token Access

kdmojdehi:
The location where you can authorize your pipeline to use system's OAuth token has changed in VSTS (it's not under options anymore)


andyjlewis:
Thanks for this pull request. We need to handle this differently because it's still true for TFS. I've got a work item on the board to track this change.

PR #1444, opened 2018-07-31
Added new line underneath version include

Jaffacakes82:
The rendering is currently broken for supported versions at the top of the page. Added a new line to allow the page to render appropriately.


alexhomer1:
Thank you for your feedback. However, I have already updated the source topic to fix this (the INCLUDE file should not be there). The updated topic should be live within 24 hours.

PR #1485, opened 2018-08-08
chore: update visualstudio.microsoft.com/get-started links

nschonni:
There are also a bunch of /build, /pipeline, and /integrate links but I think those were revereted from before


steved0x:
@nschonni Let me take this one as a private work item, we are in progress with some file structure changes and if I merged this it would cause conflicts, which often get resolved favoring the private repo by our vendors and these get lost/reverted.

PR #1502, opened 2018-08-09
Updated how to review results with parameters

ChristopherMank:
Added more verbiage around how the test outcome is generated when using parameters.


alexhomer1:
Thank you for your input!

PR #1557, opened 2018-08-19
Fix typo - "data of birth"

scovetta:
Minor typo, "data of birth" should be "date of birth".


steved0x:
@scovetta Thank you for your contribution. I am merging this PR and it should be live soon.

PR #1693, opened 2018-09-11
pipelines: Fix vs2017-win2016 image name

vbfox:
Fix vs2017-win2016 image name being incorrect in the table (but correct in the following samples)


vbfox:
Ah duplicate of #1675

PR #1715, opened 2018-09-12
cmdline task: remove displayName

paulirish:
afaik this option isn't supported by this task.
https://github.com/Microsoft/vsts-tasks/blob/master/Tasks/CmdLineV2/cmdline.ts

is that right?

### DIFF
--- a/docs/pipelines/tasks/utility/command-line.md
+++ b/docs/pipelines/tasks/utility/command-line.md
@@ -36,7 +36,6 @@ The CmdLine task also has a shortcut syntax in YAML:
 ```yaml
 - script: # script path or inline
   workingDirectory: #
-  displayName: #
   failOnStderr: #
   env:  # mapping of environment variables to add
 ```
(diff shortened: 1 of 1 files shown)


vtbassmatt:
Thanks for your contribution. `displayName` is supported on all step types in YAML, even if the task itself doesn't know anything about it. I'm not even sure we would pass it through to the task if the task asked for it.

paulirish:
okay thanks for confirming. i was unable to get this working, but i'll give it another go. 👌

vtbassmatt:
Weird, let me know what you find. [Here's](https://github.com/Microsoft/vsts-tasks/blob/17e97a41758359524ab5fd8168df0a42a4507823/ci/build-all-steps.yml#L29) an example where we use that construction in one of our own Pipelines.

paulirish:
All good! I don't remember my original attempt with `script:` but after it failed I tried this roundabout approach with:

```yaml
- task: CmdLine@2
  inputs:
    displayName: 'install yarn'
    script: 'npm install -g yarn'
```

In this case, the displayName fails: image

But anyway. At this point the easiest solution works so I'm good! :)

vtbassmatt:
Go Hokies!

And I see now... it's because displayName isn't an input. It should read:

```yaml
- task: CmdLine@2
  displayName: 'install yarn'
  inputs:
    script: 'npm install -g yarn'
```

Or in the more concise syntax we offer for the CmdLine task:

```yaml
- script: 'npm install -g yarn'
  displayName: 'install yarn'
```

PR #1735, opened 2018-09-13
Logging in to azurecr should only use the dockerId

peterj:
This line doesn't work, login fails `docker login -u $(dockerId).azurecr.io .....`. you only need to provide the `$(dockerId)` when logging in.


steved0x:
This has been fixed with a PR to our private repo. Thank you

PR #1750, opened 2018-09-14
Fix plugin name

035:


nschonni:
Think the new name is correct, just pending https://github.com/Microsoft/azure-devops-intellij/pull/174

PR #1760, opened 2018-09-15
Highlight code as JSON

DominicJ2:
The code blocks contain JSON code, but do not use syntax highlighting.  This change will make it so that the code blocks use JSON syntax highlighting.


steved0x:
@DominicJ2 Thank you for for adding that. I am merging this PR and it should be live soon.

PR #1762, opened 2018-09-16
fix broken link and use full description for term

ArieHein:
new link to Azure DevOps Services was broken and fixed to the use of developemnt instead of dev


steved0x:
@ArieHein Thank you for for fixing these. I am merging this PR and it should be live soon.

PR #1779, opened 2018-09-18
fixed variable usage

tspascoal:
One of the YAML examples was not using the right syntax to reference a variable


alexhomer1:
Thank you for your feedback.

PR #1780, opened 2018-09-18
Remove extra ::: moniker-end text

hkamel:
Remove extra ::: moniker-end text. check the image below

image


alexhomer1:
Thank you for your feedback.

PR #1781, opened 2018-09-18
Update create-project.md

Makovec:
Just a typo


steved0x:
@Makovec Thank you for for fixing this. I am merging this PR and it should be live soon.

Issue #1770, opened 2018-09-17
parallel vs. matrix


janpio:
&gt; Slicing
&gt; When `parallel` is specified and `matrix` is not defined ...

Does this mean `parallel` and `matrix` can also be used at the same time? [Multi-Configuration](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=vsts&tabs=yaml#multi-configuration) above this section only uses `matrix` together with `maxParallel`.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 54bf6c2f-95c1-943d-26ae-d7099143bccf
* Version Independent ID: 0833bb55-1e54-6953-ee69-de6d191d2ec1
* Content: [Build and release jobs in Azure Pipelines and TFS](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=vsts&tabs=yaml#multi-configuration)
* Content Source: [docs/pipelines/process/phases.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/process/phases.md)
* Product: **devops**
* GitHub Login: @alexhomer1
* Microsoft Alias: **ahomer**

alexhomer1:
Thank you for your feedback. I have passed this on to the product team for advice.

PBoraMSFT:
@janpio - no, matrix and parallel cannot be used together. 
For an example of how to use parallel for slicing, see https://docs.microsoft.com/en-us/azure/devops/pipelines/test/parallel-testing-any-test-runner?view=vsts

PBoraMSFT:
@vijayma @andyjlewis - can you help with this?

janpio:
Just stumbled over some documentation that shows all 3 in the same configuration:
> The full syntax to specify an agent job is:
```
...
  strategy:
    parallel: number
    maxParallel: number
    matrix: { string: { string: string } }
```
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=vsts&tabs=yaml

I'm pretty sure this is just not fully fleshed out all the way through all the docs. 
I appreciate you reopening this and replying.

vtbassmatt:
Fix in progress. Syntactically, all three are siblings. Semantically:

- You can use `maxParallel` alongside either of the strategies.
- The two strategies `parallel` and `matrix` are mutually exclusive, so you can only use one or the other.

[...]

PR #1792, opened 2018-09-18
Clarify exclusion between `parallel` and `matrix`

nschonni:
Related to #1770


vtbassmatt:
Hi @nschonni - thanks for this. I had to do a bit more surgery on the doc than just the one line. My fix is in flight from our internal repo.

nschonni:
@vtbassmatt no problem, but if it's going to be a few days before your stuff lands, it might be worth landing this since there are already a few issues that the hacky comment might have resolved

PR #1822, opened 2018-09-20
Semicolon is required

eakoning:
According to documentation on other places, the semicolon is required


vtbassmatt:
Hi @eakoning, thanks for your contribution. Where did you see that a semicolon is required? It's not, and I just confirmed that it works with or without the trailing semicolon. I'd like to fix the other docs if possible.

eakoning:
Hi Matt,

For instance here
https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md

Kind regards

Emiel


Op do 20 sep. 2018 om 21:28 schreef Matt Cooper <notifications@github.com>

PR #1840, opened 2018-09-21
Fixing formatting issue in compat section

joebourneMS:


steved0x:
@joebourneMS I made these changes internally and they will go out today

PR #1910, opened 2018-09-28
Name and URL correction for Team Field docs

MrHinsh:
-Updated name to be correct from "hinshel" to "hinshelwood"
-Updated blog URL from "blog.hinsherlwood.com" to "nkdagility.com"
-updated person link from "blog.hinshelwood.com" to "nkdagility.com/about-martin-hinshelwood"


steved0x:
@MrHinsh There was an issue with syncing changes from this public repo back to the private repo and then to the live doc site and so many fixes were delayed - this change should be live in a day or two, sorry it took so long.

PR #1981, opened 2018-10-05
Fix wrong link about SonarQube documentation

bellingard:
Looks like there was a bulk change from VSTS to Azure DevOps, which broke this link.


steved0x:
@bellingard Thank you for for fixing this. I am merging this PR and it should be live soon.

duncanp-sonar:
Hi @steved0x,
Could you give an update on when the change will be made live please?
Thanks!

steved0x:
@duncanp-sonar It will be live this week,.

duncanp-sonar:
@steved0x thanks for the update.

PR #1986, opened 2018-10-05
Remove duplicated installation instruction

lukamaletin:


steved0x:
@lukamaletin Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #1990, opened 2018-10-05
Minor fixes to broken links

jstoker:
Minor fix to two broken links in the following documents.
- Service hooks for Azure Storage
- How to guide, develop, add a chart


steved0x:
@jstoker Thank you for for fixing these. I am merging this PR and it should be live soon.

Issue #1980, opened 2018-10-05
it should be "APIs", not "API's"


ranjan-ranjan:
it should be "APIs", not "API's"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 294772ce-58a0-bb83-8927-3e69beca39df
* Version Independent ID: d9c4a4b3-3abf-d738-369e-fff8532e9dd9
* Content: [Guidance for authentication](https://docs.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/authentication-guidance?view=vsts)
* Content Source: [docs/integrate/get-started/authentication/authentication-guidance.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/integrate/get-started/authentication/authentication-guidance.md)
* Product: **devops**
* GitHub Login: @elbatk
* Microsoft Alias: **elbatk**

steved0x:
@ranjan-ranjan This is fixed in the associated PR, thank you.

PR #1997, opened 2018-10-07
Fix inconsistency with APIs / API's

exoen:
Fix issue #1980


steved0x:
@exoen Thank you for for fixing these. I am merging this PR and it should be live soon.

PR #2014, opened 2018-10-08
Remove en-us from links

nschonni:
Just a quick find/replace of some of the Microsoft site links to remove the language code from them


steved0x:
@nschonni I'm merging these and will watch to make sure they get through, with pre-apologies in case there is a conflict when merging back to the private repo.

PR #2048, opened 2018-10-11
Highlight that helm installer runs init client-only on install

rh072005:
I ran into an issue with helm initialisation over on the helm task ([vsts-tasks/#8315](https://github.com/Microsoft/vsts-tasks/issues/8315)).
As part of this I was told that the Helm installer task runs `helm init --client-only`.
I couldn't find reference to this in the docs (probably because it was new) but wanted to add it in


PRMerger9:

ktoliver:
Hi @shashankbarsin @azooinmyluggage 

Would you add a comment to indicate what action to take with this PR from 2018?

Merge the PR: `#sign-off`
Close the PR: `#please-close`
Wait on the PR: `#hold-off`

Thanks!

shashankbarsin:
The task runs helm init --client-only only in the case of Helm 2 and not in the case of Helm 3. Also this task is only for configuring helm on the agent and not for setting up tiller on the cluster. Closing the PR as a result

Issue #2038, opened 2018-10-10
The API URI is outdated


hjlarrea:
In the second code sample the API URI is refered to as: "https://{0}.visualstudio.com/{1}/_apis/repos/git/repositories/{2}/pullrequests/{3}/statuses?api-version=4.0-preview"

And it should be: "https://dev.azure.com/{0}/{1}/_apis/git/repositories/{2}/pullRequests/{3}/statuses?api-version=4.1-preview.1"

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 915c7aa9-0969-912b-bbc4-a2b0bf9b1533
* Version Independent ID: 6a97a5e8-8397-be98-5528-70d20e8fbed8
* Content: [Use Azure Functions to create custom branch policies - Azure Repos](https://docs.microsoft.com/en-us/azure/devops/repos/git/create-pr-status-server-with-azure-functions?view=vsts)
* Content Source: [docs/repos/git/create-pr-status-server-with-azure-functions.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/repos/git/create-pr-status-server-with-azure-functions.md)
* Product: **devops**
* GitHub Login: @steved0x
* Microsoft Alias: **yohasna**

steved0x:
@hjlarrea I just fixed the URL part yesterday, but both URL formats should work (see [Can I still use visualstudio.com?](https://docs.microsoft.com/en-us/azure/devops/user-guide/what-happened-vsts?view=vsts#can-i-still-use-visualstudiocom))

I will fix the `api-version` today as I have some other changes going out. Thanks for letting us know!

steved0x:
I just pushed the fix for this, it should be live later today. Thanks for letting us know!

hjlarrea:
Thanks guys! Glad to help!

PR #2049, opened 2018-10-11
Update visualstudio.com addresses to dev.azure.com

rh072005:
Fix for #2038


steved0x:
@rh072005 Thank you for your contribution. By a coincidence, yesterday I did a sweep of all the Azure Repos docs and made these same changes internally and they went live at some point yesterday evening or this morning, and this PR is showing a conflict because of that. I'm closing the PR, but thank you for your making it, we look forward to your future contributions!

Issue #1678, opened 2018-09-11
Specify defaults for YAML


letmaik:
Is `addToPath` true or false by default? Is `architecture` x86 or x64 by default?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: e765cfad-8ab9-443a-1171-5af81a31252f
* Version Independent ID: 8ff69d93-0128-dbeb-9d63-45266ed73020
* Content: [Use Python Version](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/use-python-version?view=vsts)
* Content Source: [docs/pipelines/tasks/tool/use-python-version.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/tool/use-python-version.md)
* Product: **devops**
* GitHub Login: @erickson-doug
* Microsoft Alias: **brcrista**

DavidStaheli:
This has been completed.

PR #2050, opened 2018-10-11
Add default values

rh072005:
Add default values as per #1678


DavidStaheli:
I approved it @steved0x.  Thanks!

PR #2080, opened 2018-10-15
chore: Remove en-us from docs.microsoft.com links

nschonni:
Looks like there is also a new validation service used by other repos check this https://docs.microsoft.com/en-us/contribute/validation-reference/locale-specific-links


steved0x:
@nschonni That service is for the GitHub repos... Unfortunately at this time we can't use it for the Azure DevOps Services docs that are hosted in Azure Repos. Let me check something and I will be back here.

nschonni:
@steved0x rebased and resolved the conflicts

PR #2099, opened 2018-10-16
Fix markdown handling of "features/*" wildcard

lsegal:
Provided markdown uses an escape for the asterisk in `features/*`, and although this seems to look fine in GitHub preview, it does not render correctly on docs.microsoft.com:

image


steved0x:
@lsegal Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #2126, opened 2018-10-17
Add "set-variables-in-scripts" file

shayki5:
Now only in the release variables page, there is a tutorial about how to use the variables on Batch/PowerShell/Shell.
It's would be great if it will also available in the Build variables page.

### DIFF
--- a/docs/pipelines/build/variables.md
+++ b/docs/pipelines/build/variables.md
@@ -31,6 +31,8 @@ This is the comprehensive list of predefined build variables.
 These variables are automatically set by the system and read-only. (The exceptions are Build.Clean and System.Debug.)
 Learn more about [working with variables](../process/variables.md).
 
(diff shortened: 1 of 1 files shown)


andyjlewis:
Thanks for the PR. We had this section in here, but it's been moved here: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=vsts&tabs=yaml%2Cbatch#set-in-script

shayki5:
Ok, I missed this page :) but why is also exist on https://docs.microsoft.com/en-us/azure/devops/pipelines/release/variables?view=vsts&tabs=batch ?

shayki5:
Ok, I missed this page :) but why is also exist on https://docs.microsoft.com/en-us/azure/devops/pipelines/release/variables?view=vsts&tabs=batch ?

andyjlewis:
@alexhomer1 might be able to advise on that question.

PR #2160, opened 2018-10-20
 Fix syntax error (publish build artifacts)

williamcspereira:


steved0x:
@williamcspereira Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #2206, opened 2018-10-24
Updated name of AccessToken variable

krmeda:
Azure DevOps uses `SYSTEM_ACCESSTOKEN` for the Access Token and not the value that was in the documentation. Fixing documentation to avoid having to find that out the hard way.


elbatk:
Thanks @krmeda !

PR #2208, opened 2018-10-24
Update manage-conditional-access.md - CAPs enforces PATs

mfjerome:
Removing warning notice about lack of support for CAPs using PATs. With sprint 142 update to Azure DevOps, CAPs are now honored with PATS, SSH keys, alternate creds and Oauth.


steved0x:
@mfjerome Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #2220, opened 2018-10-24
Update ability to run multiple agent-less tasks

alanboy:
You can now run multiple agent less tasks, so the doc needs to reflect that.


andyjlewis:
Thanks, @alanboy .

PR #2250, opened 2018-10-28
Fixed examples of script and powershell tasks.

lextm:
The original order of the two examples are wrong, so change them to match the actual section order.


steved0x:
@lextm Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #2276, opened 2018-10-30
Fix typo: synomymous -> synonymous

nightlark:
Found a small typo while I was looking for the variable to get the source checkout directory.


steved0x:
@nightlark Thank you for for fixing this. I am merging this PR and it should be live soon.

PR #2280, opened 2018-10-30
Update sprint-142-update.md

divais:


divais:
@steved0x I don't see these changes on live docs https://docs.microsoft.com/en-us/azure/devops/release-notes/2018/sprint-142-update#azure-test-plans

PR #2349, opened 2018-11-05
Update documentation on auto login and lock

jboulter11:
Update documentation on auto-login and lock to contain a second link which has the original author's plist.  The author was hosting the file but their links appear to be broken.  Including this would have saved me some time.

If you want to chat privately about this, feel free to reach out to my MSFT alias: jiboulte.


vtbassmatt:
If we want to depend on some specific .plist, we should probably host its contents ourselves rather than having users chase a trail of breadcrumbs.

jboulter11:
Sounds good to me!  I assumed there was a reason it wasn't in there already, copyright or something.  I'll make the change and keep the original link as attribution.

vtbassmatt:
@andyjlewis I'm good with this change. You would know better than me what standards we have around attribution, copyright, etc.

PR #2357, opened 2018-11-06
Add link to System.AccessToken for YAML/scripts

daveaglick:
Took me forever to find the correct docs for this, would have been very helpful to be linked from this page.


andyjlewis:
Thanks, @daveaglick

PR #2359, opened 2018-11-06
Remove Node.js from list of container requirements

MichaelSimons:
Per a conversation with @TingluoHuang, Node.js is not a container requirement.


vtbassmatt:
Already in flight from the internal repo.

PR #2384, opened 2018-11-08
update copy files for non windows agents

IsaGeoff:
Spent fucking forever trying to figure out why 0 files would copy before stumbling into the need to change the slashes.


steved0x:
@IsaGeoff Sorry that you ran into trouble, and thank you for making the contribution to fix it! I will get this reviewed ASAP.

vtbassmatt:
Not just macOS, but also Linux and BSD. It might be worth changing all the examples to forward slashes, since they'll also work on Windows.

nschonni:
Sounds like a product bug for https://github.com/Microsoft/azure-pipelines-tasks

steved0x:
This has been fixed in the docs, apologies for the long delay in looping back.

PR #2401, opened 2018-11-09
Update workflow-and-state-categories.md

vassilyl:
State category string InProgress doesn't have a space in it


KathrynEE:
@vassilyl - Thank you for the suggested changes and sorry for the slow response. I'm closing this PR as the suggested changes already exist in master.

PR #2416, opened 2018-11-12
Fixing Images paths

hkamel:
Fixing images paths, currently not visible on the article due to replacing the word VSTS with Azure DevOps which break the path.

image

image


steved0x:
@hkamel Thank you for fixing this, I have merged in to our private repo, and it should be live very soon. Thank you.

PR #2423, opened 2018-11-13
Fixing broken image links

hkamel:
fixing broken image links


steved0x:
@hkamel I fixed these too, thanks for letting us know.

PR #2656, opened 2018-12-05
Update transforms-variable-substitution.md

uniquelau:
Added sentence the re-enforces that the parameters.xml file should be used to handle updating values that are not within the `applicationSettings`, `appSettings`, `connectionStrings`, and `configSections` elements. Added information that indicates that variables are not updated within parameters.xml file, and has to be handled using 3rd party task such as "Replace Tokens".

**Context**

I'm currently migrating some pipelines from Octopus Deploy into Azure Pipelines. I've spent the best part of a day trying to update some values inside my applications configuration files. 

I am trying to update an element to use a variable that is set in the deployment pipeline, and update the following element `/configuration/system.net/mailSettings/smtp/network/@enableSsl`

### DIFF
--- a/release-notes/index.md
+++ b/release-notes/index.md
@@ -93,7 +93,7 @@ Versions in the “Server” column are linked to the appropriate download locat
     </thead>
     <tbody>
         <tr>
-            <td rowspan="12">[04 December 2018](2018/sprint-143-update.md)</td>
+            <td rowspan="12">[04 December 2018](2018/sprint-144-update.md)</td>
             <td>Link GitHub commits and pull requests to Azure Boards work items</td><td>Future</td>
         </tr>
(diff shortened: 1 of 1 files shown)


steved0x:
@uniquelau Thank you for your contribution. We have had some trouble with syncing between this public repo and our private repo, which has been resolved, but has caused an issue with PRs submitted during the interval in which the problem existed. I tried to clear it up, and it didn't work... Could you submit a new PR made from a branch off the latest master, or if you let me know the changes, I can make them. Sorry for the trouble, and thank you again for the contribution.

steved0x:
@uniquelau I made these changes to our internal repo and they should be live soon. Thank you for your contribution.

PR #2669, opened 2018-12-06
Fix link for 04 December 2018 to sprint 144

elanderson:


steved0x:
@elanderson Thank you for for fixing this. I am merging this PR and it should be live soon.

elanderson:
I just checked the site again and this change hasn't hit live yet. Is that normal?

PR #2714, opened 2018-12-09
Changed Stackoverflow's URL

nguyenlamlll:
After VSTS is changed to Azure DevOps, Stackoverflow has changed the tag name into `azure-devops`. I think we need to update this change.

This PR is a re-open one after syncing fixes (previously opened in #2640)


steved0x:
@nguyenlamlll Thank you for this fix, I am merging it now.

PR #2732, opened 2018-12-12
Fixing service names

buckh:
updated service names


steved0x:
@buckh Thank you for fixing this, I am merging it now and it should be live shortly.

buckh:
Thanks

Buck

Get Outlook<https://aka.ms/qtex0l> for iOS
________________________________

PR #2753, opened 2018-12-13
Added a new artifact variable

ashokirla:
Based on the customer feedback, we have introduced a new artifact variable to extract the alias for primary artifact source as a variable. https://developercommunity.visualstudio.com/content/problem/255744/release-definition-artifact-source-alias-not-avail.html


steved0x:
@ashokirla Thank you for your contribution, I have assigned this to the author of the article for review.

PR #2790, opened 2018-12-17
chore: Fix merge conflicts from 2019 moniker

nschonni:


steved0x:
@nschonni Uggg I'll get this merged and then when it comes over to the private side I'll get these reviewed to make sure we've got the correct versions in there. Thank you.

PR #2838, opened 2018-12-20
Groupnames wrong!

chrizzo84:
Think the name of the groups are "Endpoint Administrators" and "Endpoint Creators" instead of "Service connection administrators" and "Service connection creators".


alexhomer1:
Thank you for your input. I have modified the source documents and so I cannot complete this PR. Instead, I will make sure that the source documents are suitably updated. Note that "Service endpoints" are now called "Service connections" and so it's likely that the names of the roles will change at some point.

PR #2853, opened 2018-12-21
Turn off Windows Firewall

ChristopherMank:
After running into issues with a large TFS import, Microsoft support indicated to turn off the Windows Firewall. After doing so, the import continued successfully. This update is to add verbiage to the docs to perform this step.

### DIFF
--- a/docs/articles/migration-import.md
+++ b/docs/articles/migration-import.md
@@ -464,6 +464,10 @@ Set the source to "IP Addresses", enter one of the IPs that need to be granted a
 
 You will need to repeat adding inbound port rules until all necessary Azure DevOps Services IPs have been granted an exception. Missing one IP could result in your import failing to start. 
 
+#### Windows Firewall
+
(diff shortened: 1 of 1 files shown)


steved0x:
@ChristopherMank I'm sorry that you ran into trouble with your TFS Import and I'm glad that support was able to offer a workaround. I spoke with one of the engineers for TFS Import, and he doesn't feel that turning off the Windows Firewall is a recommended approach that needs to be added to the documentation. Can you share any additional details about your scenario?

ChristopherMank:
@steved0x This was for a large, 600GB TFS DB migration. Due to the size, we had to use an Azure SQL VM. I spun up a SQL VM from the Azure Gallery and configured the network settings as shown in the screenshot. This was following the documentation we are referencing to open ports only to the migrator service. I then went to run the import but it would fail saying it could not connect to the DB. I validated many times the IPs were correct and that the user account created to grant permissions was correct. After speaking with Microsoft support, they indicated to turn off Windows Firewall on VM. Once I did that, the import started successfully without doing any other step. Perhaps something else fixed this issue and it really wasn't the firewall?

net

steved0x:
@ChristopherMank Thank you for this and your other contributions. The PM for this area doesn't want to add the firewall instructions, and I am closing this PR. Sorry that it took so long for a resolution.

PR #2858, opened 2018-12-21
fix: Remove shorttitle frontmatter

nschonni:
That frontmatter doesn't appear to be used in any other articles


steved0x:
@nschonni I'm not sure what that is either, checking with the author in case it is something special.

PR #2876, opened 2018-12-28
Remove Code Coverage tab from screenshot

lupino3:
The screenshot may induce users to think that the code coverage results
are always browsable in a separate tab once they are uploaded, but this
is not the case: this feature is not available.


PBoraMSFT:
The feature is not available to anonymous users. It can be seen by users that are authenticated.
Additionally, the documentation does not talk about the code coverage tab at all. It only talks about the artifacts from the build summary page, which should be available for ALL users.

lupino3:
Thanks for the clarification! Would it be ok if I updated the documentation
with your explanation? I still find it a bit confusing that the screenshot
shows the tab but the documentation does not mention it. Thanks!

steved0x:
@PBoraMSFT I assigned this to you, please merge when it is ready, or let me know when it is ready and I can merge it.

PR #2880, opened 2018-12-29
correct capitalization

compnerd:
CMake was incorrectly capitalised, correct it.


compnerd:
CC: @andyjlewis

steved0x:
@compnerd Thank you for fixing that.

PR #2884, opened 2018-12-31
Clone of Hosted XML to Inheritance Now Exists

ChristopherMank:
Previous documentation indicated that a process would be coming to clone a Hosted XML to an Inheritance process. This clone operation now exists and updated QA section to reflect that.


nschonni:
Minor, but the links shouldn't contain the `en-us`, since the should redirect to the user's locale

ChristopherMank:
Thanks @nschonni , removed that from URLs.

steved0x:
@ChristopherMank @nschonni Thank you for your contribution!

PR #2897, opened 2019-01-03
chore: Renormalize line endings

nschonni:
Probably easier to do this on the internal side right before you do a repo sync since it might interfere with old open PRs.
I just ran `git add --renormalize .` then checked in the changed files.


steved0x:
@nschonni I have created a work item to do this internally, you should see it soon.

PR #2948, opened 2019-01-07
chore: Remove `#r "Newtonsoft.Json"`

nschonni:
Maybe there is a reason for this, but it doesn't seem to match the enclosing csharp code fence.
There is already a `using Newtonsoft.Json;` below these lines


steved0x:
@nschonni Agreed, merging. Not sure how that got in there.

PR #3008, opened 2019-01-13
Added categories and updated VSS SDK paths

jlattimer:
Categories is required in the manifest and without it you won't be able to upload the VSIX. Updated the VSS SDK file paths so the files will be dropped into the correct folder (according to the HTML) when the extension is created.


ktoliver:
The updates that are requested in the PR are reflected in master. We'll close the PR.

PR #3047, opened 2019-01-17
Fix displayName in docker login bash block

chwarr:
The displayName key needs to be at the same level as the bash key,
otherwise it is considered part of the command to execute.


alexhomer1:
Thank you for your input. However, the source topic has been rewritten and so these updates are no longer valid.

PR #3092, opened 2019-01-22
Update service-containers.md

markjgardner:
Unit tests should never require external dependencies. Integration tests, on the other hand, do.


alexhomer1:
Thank you for your input!

PR #3153, opened 2019-01-28
Added details about two new extensibility points in the Build > Test Results Tab

rishavsharan:
We have added two new contribution points in the Test view for Build;

1. New custom tab: Using this contribution point, users can create an extension which adds a new tab to the Test resultants Details Pane which comes up on selecting any data grid row.
2. New Action in the toolbar: Using this contribution point, users can create extension which add custom action buttons in the Test results view toolbar and the corresponding action on the click event.


rishavsharan:
Closing in favor of the internal process.

PR #3233, opened 2019-02-05
powershell arguments fix

damccorm:
Make it clear that arguments are ignored when using an inline script

Fixes https://github.com/Microsoft/azure-pipelines-tasks/issues/9403


steved0x:
@damccorm Thank you for your contribution.

PR #3286, opened 2019-02-07
Fix wrong word

letmaik:


steved0x:
@letmaik Thank you for fixing that.

PR #3371, opened 2019-02-14
use-ssh-keys-to-authenticate.md: list actual hosted fingerprints

congyiwu:
Officially document the fingerprints for the SSH host key used across all of hosted Azure DevOps.

Currently the fingerprints were only visible after signing in to a specific organization.

@sarath-p requested public documentation that can be referenced from non-organization specific projects, e.g. https://github.com/weaveworks/flux/blob/master/docker/verify_known_hosts.sh


PRMerger6:
@congyiwu : Thanks for your contribution! The author, @steved0x, has been notified to review your proposed change.

PR #3388, opened 2019-02-15
fix: Change from bolded NOTE to block NOTE

nschonni:
Not sure if it should be a different alert type, but left it as a "note"


PRMerger18:
@nschonni : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

PR #3404, opened 2019-02-18
Edited private-powershell-library.md

istvan2015:
Edited private-powershell-library.md to include example for <org>.pkgs.visualstudio.com (PS code example)

### DIFF
--- a/docs/artifacts/tutorials/private-powershell-library.md
+++ b/docs/artifacts/tutorials/private-powershell-library.md
@@ -223,7 +223,7 @@ We now have a private repository within Azure Artifacts that we can push our Pow
     
     > **NOTE:** You will notice above that the Publish and Source location both reference Version 2 of NuGet. PowerShell does not support Version 3 of NuGet.
     
-    Another example for Azure Artifacts, which defaults to Version 3 of NuGet (but you will use Version 2):
+    If you're still using the older ```visualstudio.com``` URLs, use this command instead:
 
     ```powershell
     Register-PSRepository -Name "PowershellAzureDevopsServices" -SourceLocation "https://<org_name>.pkgs.visualstudio.com/_packaging/<feed_name>/nuget/v2" -PublishLocation "https://<org_name>.pkgs.visualstudio.com/_packaging/<feed_name>/nuget/v2" -InstallationPolicy Trusted
--- a/docs/artifacts/tutorials/private-powershell-library.md
+++ b/docs/artifacts/tutorials/private-powershell-library.md
@@ -202,6 +202,11 @@ We now have the module and the module manifest. We are ready to package it and s
     ```powershell
     nuget.exe sources Add -Name "PowershellModules" -Source "https://pkgs.dev.azure.com/<org_name>/_packaging/<feed_name>/nuget/v3/index.json"
     ```
+    
+    If you're still using the older ```visualstudio.com``` URLs, use this command instead:
+    ```powershell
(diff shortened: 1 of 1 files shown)


PRMerger18:
@istvan2015 : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

PRMerger20:
@istvan2015 : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

elbatk:
Reviewing this soon! Thanks @istvan2015

istvan2015:
Great, thank you. Let me know what you think

docs/artifacts/tutorials/private-powershell-library.md
```diff
@@ -179,6 +222,12 @@ We now have a private repository within Azure Artifacts that we can push our Pow
     ```
     
     > **NOTE:** You will notice above that the Publish and Source location both reference Version 2 of NuGet. PowerShell does not support Version 3 of NuGet.
+    
+    Another example for Azure Artifacts, which defaults to Version 3 of NuGet (but you will use Version 2):
```
alexmullans: Can we change this to say more about when it would be used? e.g. "If you're still using the older visualstudio.com URLs, use this command instead:"

PRMerger13:
@istvan2015 : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

docs/artifacts/tutorials/private-powershell-library.md
istvan2015: Thank you, nice observation! I changed it. You're right that my wording wasn't specific enough for the scope of the sample URLs provided

nschonni:
Outside of what you changed, but I think https://github.com/MicrosoftDocs/vsts-docs/blame/02752afb793a1583b7464ac4ce9c79211e2a90d4/docs/artifacts/tutorials/private-powershell-library.md#L203 needs to be switched, since it is also referencing the v3 endpoint

istvan2015:
@nschonni Thank you for the input. As far as I am aware and as far as I have tested, that is not true, because I was able to successfully connect NuGet using the v3 endpoint - which is actually whichever endpoint the developers have to use for their current Azure DevOps setup.

However, your comment did open my eyes to the fact that the line you mentioned is correct, but I need to add another line below to include the example URL for the old ```visualstudio.com``` versions as well, exactly as @alexmullans pointed out.

I hope that makes sense

nschonni:
PowerShellGallery doesn't support NuGet v3, only v2. That is mentioned in other parts of the article

nschonni:
I might be mixing up how this particular v3 endpoint is being used. Maybe publishing to v3 is fine, as long as the pull is done from v2

istvan2015:
> I might be mixing up how this particular v3 endpoint is being used. Maybe publishing to v3 is fine, as long as the pull is done from v2

Well, you are right in the sense that PowerShellGallery doesn't support Nuget v3. However, that particular line (line # 203) only adds a feed to Nuget locally on the computer from which it was called, and names it ```PowershellModules```. That's why it works with v3, because it's simply adding a feed, you give it a name, of course it works.

Sorry if I am mistaking, I'm not trying to enforce my point of view here, just trying to explain my limited experience with these bits.

PRMerger18:
@istvan2015 : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

alexmullans:
@istvan2015 thanks for the contribution, we appreciate it!

PR #3420, opened 2019-02-19
Adding examples for github releases task

raiyanalam:
Added examples for create, edit and delete scenarios.


DavidStaheli:
# So helpful!  👍

PR #3478, opened 2019-02-25
Clarify that Agent.WorkFolder may not be writable

flcdrg:
- Especially true when folder is mapped into a container

As mentioned by @TingluoHuang in https://github.com/Microsoft/azure-pipelines-agent/issues/2122#issuecomment-467024271


flcdrg:
I didn't update the documentation for earlier TFS versions. I can do that to make it consistent, but figured it may not be necessary as I don't think those editions supported containers, so technically you could get away with writing to WorkFolder if you tried.

PR #3592, opened 2019-03-08
AngularJS is v1. Angular is the correct term for v2+

johnpapa:
AngularJS is v1. Angular is the correct term for v2+


PRMerger9:
@johnpapa : Thanks for your contribution! The author, @andyjlewis, has been notified to review your proposed change.

andyjlewis:
Thanks, @johnpapa

Issue #3513, opened 2019-02-28
Invalid YAML in trigger documentation


nickbabcock:
The documentation lists (for tags and branches)

```yaml
trigger:
  tags:
    include:
    - *
```

Is invalid yaml. You're probably looking for:

```yaml
trigger:
  tags:
    include: ['*']
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: f5673ec1-c7d1-e0ac-043a-8fc3d5b226a9
* Version Independent ID: b4872831-51c0-6cf8-1c56-de20beb38a80
* Content: [Build pipeline triggers - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml)
* Content Source: [docs/pipelines/build/triggers.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/build/triggers.md)
* Product: **devops**
* GitHub Login: @steved0x
* Microsoft Alias: **sdanie**

ryanmaclean:
@steved0x let me know if that's OK, ran into the same thing and since it was already reported and pretty easy, figured I'd help out.

piotr-napadlek:
Thanks man, I just spent like an hour on a basic task of creating a pipeline for all non master branches, thanks to misleading docs I was hitting the wall until I read your post and tried:
```trigger:
  branches:
    exclude: ["master"]
    include: ["*"]```

**Comment by steved0x** [comment]
@nickbabcock @ryanmaclean Sorry that you ran into trouble, we have got this updated in our docs with a slightly different syntax:

```
trigger:
- master
- releases/*
```

and:

```
trigger:
  branches:
    include:
    - '*'  # must quote since "*" is a YAML reserved character; we want a string
```

I think I remember seeing this PR come in to the public repo, and it conflicted with a nearly simultaneous [PR to the private repo](https://github.com/MicrosoftDocs/vsts-docs/commit/20580ae9076d549cc983d36b2b683724daf06502) (possibly after seeing this issue?) and the merge conflict was resolved in favor of the private repo.

Sorry for the trouble and for the great delay in getting back to this GH issue, and thank you for letting us know.

PR #3613, opened 2019-03-12
Addressing YAML Typo from #3513

ryanmaclean:
The YAML in the example should be changed to `'[*]'` as noticed in issue #3513.


PRMerger19:
@ryanmaclean : Thanks for your contribution! The author, @steved0x, has been notified to review your proposed change.

vtbassmatt:
Good catch. Because we scan all values as strings (and don't support every YAML feature), this is actually valid in our parser. A bare "*" in a sequence gets treated as a string with or without the single-quotes. I don't care for the inline-sequence style, though, and would prefer to leave the example as-is.

PR #3757, opened 2019-03-25
Add missing part in the Git users command

ShaneZhengNZ:
Also add notes to fix the issue when 'authors-transform.txt' file encoding is not correct and author is not found.


PRMerger18:
@ShaneZhengNZ : Thanks for your contribution! The author, @hkamel, has been notified to review your proposed change.

ktoliver:
The update that's requested in the PR appears to be reflected in master. We'll close the pull request.

PR #3761, opened 2019-03-25
Simple approach not supported in AzureDevOpsServer

RoopeshNair:
Simple approach is not supported in Azure DevOps Server 2019 or earlier versions of TFS. User needs to use advanced version only


PRMerger6:
@RoopeshNair : Thanks for your contribution! The author, @alexhomer1, has been notified to review your proposed change.

alexhomer1 (approved):

PR #3824, opened 2019-04-01
Misleading registry name results in access denied

SvenPam:
As highlighted in this Stackoverflow question: https://stackoverflow.com/questions/55421635/resource-access-denied-when-pushing-container-to-azure-container-registry/55450990, Azure DevOps pipeline appears to default to the docker hub when just the namespace is provided. I've updated the example to illustrate that using the fully qualified ACR registry name (as provided from Azure) resolves the issue.


PRMerger6:
@SvenPam : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x:
Sorry for the long delay, this topic doesn't exist any longer, and its content has been split out among the topics in this node:

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/build-image?view=azure-devops&tabs=java

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/acr-template?view=azure-devops&tabs=java

I wasn't able to find the lines that you fixed there so it may have been fixed already during the content refactor. Thank you for your contribution

PR #3840, opened 2019-04-02
Fix environment variable reference.

lb5tr:
Current example does not work with ADO pipelines.

### DIFF
--- a/docs/pipelines/languages/docker.md
+++ b/docs/pipelines/languages/docker.md
@@ -230,10 +230,10 @@ variables:
 steps:
 - script: |
     docker build -t $(dockerId)/$(imageName) .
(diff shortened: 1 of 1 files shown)


PRMerger7:
@lb5tr : Thanks for your contribution! The author, @andyjlewis, has been notified to review your proposed change.

steved0x:
@lb5tr Sorry for the long delay, this topic doesn't exist any longer, and its content has been split out among the topics in this node:

https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/build-image?view=azure-devops&tabs=java

I wasn't able to find the lines that you fixed there so it may have been fixed already during the content refactor. Thank you for your contribution

Issue #1580, opened 2018-08-21
Example content for testAssemblyVer2 needs to express newlines


jimmylewis:
There needs to be newlines to separate the match expressions.  This could be done with double quote and escape characters, e.g: 
`"**\\*test*.dll\n!**\\*TestAdapter.dll\n!**\\obj\\**"`

or by making use of multiline operators in YAML, e.g.:

```
#testAssemblyVer2: |
#   **\*test*.dll
#   !**\*TestAdapter.dll
#   !**\obj\**
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 16bd8dd1-7650-8c18-2553-38a8bc54e892
* Version Independent ID: f9fa517a-7adc-e327-d12e-256295a1ed52
* Content: [Visual Studio Test](https://docs.microsoft.com/en-us/vsts/pipelines/tasks/test/vstest?view=vsts)
* Content Source: [docs/pipelines/tasks/test/vstest.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/test/vstest.md)
* Product: **devops**
* GitHub Login: @erickson-doug
* Microsoft Alias: **dastahel**

jimmylewis:
bumping this to give it a little more lifetime, as the existing example does not work - users who copy it will have to debug the syntax issues.

tillig:
This just ate my lunch for a couple of hours. 😠 Getting the docs correct and using _double quotes_ in the example to ensure `\n` gets properly interpreted is pretty important to getting people on the path to success, especially when most of the editing is done in YAML instead of through a nice visual designer that handles the serialization.

It wouldn't hurt to get some YAML multiline examples [on the pattern matching page, too](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/file-matching-patterns?view=azure-devops&viewFallbackFrom=vsts) since none of the ones there would actually work if you pasted them into a pipeline YAML doc.

I'm still fighting through trying to get multiple exclusions (`!pattern`) working, starting to think it's related to #2551. Anyway, having a working example would be great.

TroySchmidt:
Definitely need to fix this.  Even using the new online YAML graphical editing puts in invalid values for `testAssemblyVer2` and *ONLY* by seeing this thread was I able to use the pipe multi line and fix it.

prawalagarwal:
[...]

PR #3953, opened 2019-04-13
Use a valid example for the testAssemblyVer2 property

jimmylewis:
The existing sample is not valid and won't work.  Multiple users have
wasted time with a bad example.

This addresses #1580


PRMerger9:
@jimmylewis : Thanks for your contribution! The author, @, has been notified to review your proposed change.

prawalagarwal:
@jimmylewis Do you have any thread going with Andy on this? What are the updates here.

steved0x:
Fixed in another PR

PR #3982, opened 2019-04-16
Use native PowerShell throughout example

SQLDBAWithABeard:
Just a change to use native PowerShell throughout the example code instead of mixing and matching


PRMerger6:
@SQLDBAWithABeard : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #3983, opened 2019-04-16
markdown/images/resizing: Remove note about resizing only working in certain contexts

juharris:
Seems like resizing does now work when looking at a README when browsing the repo.


PRMerger10:
@juharris : Thanks for your contribution! The author, @chcomley, has been notified to review your proposed change.

PR #4061, opened 2019-04-24
Un-comment AzureKeyVault@1 secretsFilter argument

c-w:
According to the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/deploy/azure-key-vault?view=azure-devops#arguments), the `secretsFilter` argument is required. As such, this change un-comments the argument in the snippet to avoid potential confusion for users since having the argument commented-out may be interpreted as the argument being optional.


PRMerger16:
@c-w : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #4067, opened 2019-04-25
Add newlines between match expressions in testAssemblyVer2

dlumpp:
The example for testing using `VSTest@2` assembly pattern matching expressions contains 3 expressions with no separation. If you try to copy and paste the sample into your own YAML it won't work correctly. 

Had it not been for issue #1580 clarifying this, I would have wasted a lot of time figuring this out. 

This PR adds newline characters to separate the example expressions.


PRMerger15:
@dlumpp : Thanks for your contribution! The author, @, has been notified to review your proposed change.

PR #4095, opened 2019-04-30
Publish link was linking to Create Publisher

remcoeissing:
Publish link was linking to Create Publisher. Instead of linking to the manage extensions page, manage extension will redirect to create publisher if the account is not a publisher.


PRMerger10:
@remcoeissing : Thanks for your contribution! The author, @elbatk, has been notified to review your proposed change.

PR #4097, opened 2019-04-30
Remove A: from Q&A section

remcoeissing:
Only in the answer on "I need more agents. What can I do?" was an "A: ". Removed it for consistency


PRMerger17:
@remcoeissing : Thanks for your contribution! The author, @andyjlewis, has been notified to review your proposed change.

PR #4101, opened 2019-04-30
escaped colons

yuhattor:
colons are needed to be escaped in yml.


PRMerger17:
@yuhattor : Thanks for your contribution! The author, @, has been notified to review your proposed change.

steved0x:
@yuhattor Thank you for fixing this. I am merging this, but I don't think this particular file is used anywhere, which is why we have't gotten any doc build warnings. I will look into this and we may end up deleting this file.

PR #4214, opened 2019-05-13
Clarified how variables from variable groups should be used in YAML f…

IgorPietraszko:
When reviewing the documentation, it was unclear to me how variables from variable groups should be used in a YAML script.  I hope that the proposed changes clarify this topic for any other users.


PRMerger12:
@IgorPietraszko : Thanks for your contribution! The author, @RoopeshNair, has been notified to review your proposed change.

PR #4224, opened 2019-05-15
Update bash.md

kejxu:
https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/bash?view=azure-devops

meantime, the following code block does not render correctly:
```yaml
- script: echo $MYSECRET
  env:
    MySecret: $(Foo)
```

not sure if there is a global styling .css file, maybe try something like this instead:
```
<code style=display:block;white-space:pre-wrap>yaml
- script: echo $MYSECRET
  env:
    MySecret: $(Foo)</code>
```

### DIFF
--- a/docs/pipelines/tasks/utility/bash.md
+++ b/docs/pipelines/tasks/utility/bash.md
@@ -47,7 +47,7 @@ The Bash task also has a shortcut syntax in YAML:
 <td>If this is <code>true</code>, this task will fail if any errors are written to <code>stderr</code>.</td>
 </tr>
 <tr>
-<td>Env[ironment variables]</td>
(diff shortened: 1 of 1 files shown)


PRMerger10:
@kejxu : Thanks for your contribution! The author, @vtbassmatt, has been notified to review your proposed change.

vtbassmatt:
Thanks. I'm not sure why someone styled "Environment variables" that way. I will also fix the code block -- good catch!

nschonni:
> Thanks. I'm not sure why someone styled "Environment variables" that way

I think it's because the actual YAML key is `env`. Other docs use this style for shorthand/longhand with commandline options

PR #4290, opened 2019-05-21
Fix broken link about PR Research

spottedmahn:
The [current link](https://www.microsoft.com/research/publication/convergent-software-peer-review-practices/) doesn't work.  I don't know if defaulting to `en-us` is the right approach but it fixes the link 😊

### DIFF
--- a/docs/repos/git/git-branching-guidance.md
+++ b/docs/repos/git/git-branching-guidance.md
@@ -83,7 +83,7 @@ Distribute reviewer responsibilities to share ideas across your team and spread 
 
 Some suggestions for successful pull requests:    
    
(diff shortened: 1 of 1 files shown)


PRMerger14:
@spottedmahn : Thanks for your contribution! The author, @apawast, has been notified to review your proposed change.

steved0x:
@spottedmahn Thank you for making this fix. The original link works for me, and if we add in the `en-us` it will be taken out the next time we do a sweep for locale information which usually shouldn't be hard-coded into links. Please let us know if it keep happening. Thank you.

spottedmahn:
Works for me too, today that is. 🤔

PR #4386, opened 2019-05-30
Fix group full name slashes

chwarr:
Groups are separated from projects by a backslash. Forward slashes do
not work in the Azure Dev Ops web UI in all places.

Backslashes are sometimes used to escape other Markdown characters, so
they're doubled here. They render as a single backslash.

### DIFF
--- a/docs/organizations/security/permissions.md
+++ b/docs/organizations/security/permissions.md
@@ -225,9 +225,9 @@ You can't remove or delete the default server level groups.
 </tbody>
 </table>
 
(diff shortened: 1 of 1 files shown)


PRMerger8:
@chwarr : Thanks for your contribution! The author, @KathrynEE, has been notified to review your proposed change.

PRMerger8:
@chwarr : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chwarr:
Rebased against the current master commit, c0330b9a1.

PR #4434, opened 2019-06-05
Updating URL samples to new format in npmrc.md

MartynJones87:
The previous URLs used in the Create a token that lasts longer than 90 days section were using the old format of URL for the artifacts NPM feeds, not the new style that are now being used. This caused me to waste a long time debugging an authentication error as I didn't notice the difference in URL format.


PRMerger6:
@MartynJones87 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MartynJones87:
Sorry, I've just realised, after some research by my colleague, that the URL's in the documentation are already in the newer format and our Organization was using the legacy URL's. I will close this PR, as it is wrong.

Issue #4612, opened 2019-06-24
Not just pushing but also restoring from a TFS Package Management feed with basic auth enabled does not work.


Ausm:
The statement here https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget?view=azure-devops#publishing-packages-to-tfs-with-iis-basic-authentication-enabled states that pushing to a TFS Package Management feed does not work when basic auth is enabled on the TFS.
This is correct, but restoring from such a feed also does not work since it needs to get the feeds URI and for that to be authenticated against the TFS, which does not work when basic authentication is switched on on the TFS.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 0fa37a3f-6551-16ad-e164-a4ac522cdaab
* Version Independent ID: a3fd6935-2b99-92ef-4eb3-3249fca01f8a
* Content: [NuGet restore, pack, and publish task - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget?view=azure-devops#push-nuget-packages)
* Content Source: [docs/pipelines/tasks/package/nuget.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/package/nuget.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @elbatk
* Microsoft Alias: **elbatk**

elbatk:
Hey there @Ausm , I've approved and merged your PR. Thank you for the contribution!

PR #4614, opened 2019-06-24
Add restoring to basic auth statement

Ausm:
Add that not just publishing but also restoring may fail if basic authentication is enabled on the TFS as described in #4612.


PRMerger16:
@Ausm : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #4657, opened 2019-06-27
Update xamarin.md

auroraloop:
We're proposing this change to the docs because we recently had a hard time configuring the Xamarin.iOS build task to produce an .ipa package for CI/CD in Azure Pipelines. We had to troubleshoot a lot and we even had to perform the build using a regular bash task to finally find out that we were passing a wrong apple certificate. The Xamarin.iOS task never through an error but we noticed no .ipa file was being generated at all by carefully reviewing the folders of the agent, somehow a regular bash msbuild command provided more information pointing we were missing signing identities for the app bundle id.

After understanding the issue we guessed it be good idea to share this with our customer through this docs.
Hopefully, you'll find our contribution useful. Proposing this with @ManuelGalindo we're both Azure DevOps support engineers. Pending for corrections if needed we can provide a visual guide with screenshots for the classic pipeline tab.

### DIFF
--- a/docs/pipelines/languages/xamarin.md
+++ b/docs/pipelines/languages/xamarin.md
@@ -71,7 +71,7 @@ steps:
 
 ### Sign a Xamarin.Android app
 
-See [Sign your mobile app during CI](../apps/mobile/app-signing.md) for information about signing your app.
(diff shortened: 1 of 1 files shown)


arammaliachi:
@steved0x @madhurig could you take a look at this? :)

PR #4769, opened 2019-07-07
fix: MD035/hr-style

nschonni:
Consistent horizontal rule style. Could also probably be removed.


nschonni:
Rebased an updated rule to keep a consistent style

PR #4806, opened 2019-07-10
Fixed password variable settings.

yuhattor:
$(pswd) didn't work. It should be $pswd
I suppose these are mistakes.
Should change from $(pswd) to $pswd, or from $(pswd) to $(dockerPassword) because variable pswd is defined on env section.


PRMerger8:
@yuhattor : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
The documentation has been updated and the article no longer exists. We'll close the PR.

Issue #4980, opened 2019-07-23
Missing description for externalFeedCredentials


sid-dahiya:
On page: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops, there is no documentation for `externalFeedCredentials` attribute. This YAML parameter is extremely useful when connecting to multiple authenticated feeds via Service Connections.

For reference, this string may contain a comma separated list of Service Connections such as:
`externalFeedCredentials: 'SC1,SC2,SC3,...,SCn'`

sid-dahiya:
Sent out PR #4981 to address this.

cxwtool:
This issue hasn't been updated in more than 180 days, so we've closed it. If you feel the issue is still relevant and needs fixed, please reopen it and we'll take another look. We appreciate your feedback and apologize for any inconvenience.

PR #4981, opened 2019-07-23
Update dotnet-core-cli

sid-dahiya:
Added `externalFeedCredentials` description as that is in the sample on top, but not described. Also added example with multiple values.

**Question for maintainers**: Since `externalEndpoints` and `externalFeedCredentials` are supposed to be same, as per my understanding, do we want to have two separate entries for them?


sid-dahiya:
This is related to #4980

PRMerger8:
@sid-dahiya : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
The version of the article in master appears to address the issue raised in the pull request. We'll close the PR.

PR #5125, opened 2019-08-03
Update azure-resource-group-deployment.md

hintbw:
It is very unclear in the documentation what value should be included here. Especially since the value to be provide should be the GUID/resourceID of the service connection created for the Azure subscription being used. I propose to add a little more text to that particular field to indicate that.


hintbw:
I find that my pull request is probably inaccurate here after reading http://donovanbrown.com/post/How-to-target-a-different-Azure-subscription-without-changing-every-task and the values he is able to provide. It seems like there are several valid approaches to what to pass to this property, but there is no documentation around what those possible values could be. My PR is likely incorrect here as a result

PRMerger9:
@hintbw : Thanks for your contribution! The author(s) have been notified to review your proposed change.

RoopeshNair:
@hintbw  - closing the PR per your comments. Feedback taken on improving the doc

Issue #5150, opened 2019-08-06
Add DotNetCoreCLI task to list of tasks that auto publish results


patrickhuber:
using the following pipeline:

```
trigger:
- master

pool:
  vmImage: 'ubuntu-latest'

steps:
- task: DotNetCoreInstaller@0
  displayName: 'Install .net core 3.0 (preview)'
  inputs:
    version: '3.0.100-preview6-012264'

- task: DotNetCoreCLI@2
  displayName: Build
  inputs:  
    command: 'build'
    projects: |
      **/*.sln

# uncomment to display all output files
#- powershell: dir -Recurse
  
- task: DotNetCoreCLI@2
  displayName: Test
  inputs:
    command: 'test'
    projects: |
      **/*Tests*.csproj
```

The test results are automatically published. I assume this is due to the DotNetCoreCLI@2 task auto publishing the results. I don't see this in the documentation though. 

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: ba91e46a-de53-c0c1-f7f7-c02e797d527c
* Version Independent ID: 6f478148-461f-694a-ed82-deb4d731854e
* Content: [Publish Test Results task - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml#feedback)
* Content Source: [docs/pipelines/tasks/test/publish-test-results.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/test/publish-test-results.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @PBoraMSFT
* Microsoft Alias: **pbora**

divais:
@patrickhuber the details are available in the Don't net core CLI task - https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#test

PR #5151, opened 2019-08-06
Update publish-test-results.md

patrickhuber:
Adds dotnet core cli as a task that auto publishes. 

See issue #5150


PBoraMSFT:
@divais - FYA as necessary.

PR #5178, opened 2019-08-08
Update feed locator URL

MikeBazMSFT:
The documentation at https://docs.microsoft.com/en-us/azure/devops/artifacts/how-to/set-up-upstream-sources?view=azure-devops indicates a different feed locator URL format than indicated here (a different protocol is referenced.)  This change updates this page to match the other page, and clarifies the header for that area.


chasewilson:
@MikeBazMSFT Thanks for the contribution :) Merging and closing

PR #5185, opened 2019-08-08
Template expansion can only use system variables

kzu:
Clarifies for others facing https://github.com/microsoft/azure-pipelines-yaml/issues/30 or https://github.com/microsoft/azure-pipelines-yaml/issues/297


MisinformedDNA:
I think the documentation should explicitly say what variables are _not_ allowed.

PR #5202, opened 2019-08-09
just fixing a typo

michael-hamilton:


chasewilson:
@michael-hamilton, thanks for your contribution:)

PR #5204, opened 2019-08-11
Adding AzureRM commands

wviriya:
Adding AzureRM sample. The only sample provides is for ASM. I could not get the script to check for error. It seems that the ErrorAction switch does not work with AzureRM.Sql. I want to also add Az module, but it will only work with the preview version 4 of Azure PowerShell task.


Ajay-MS:
@wviriya  Thanks for your contribution for the samples.

Ajay-MS:
@RoopeshNair  - I have approved the PR. Can you also please review and merge the PR.

PR #5305, opened 2019-08-20
Fixed casing of allowed parameter values

NenoLoje:
The parameters for `deploymentMode `should all start with capital letters (otherwise an error occurs).


NenoLoje:
To add to this: This PR is only changing comments (as the parameter values were cased incorrectly).

NenoLoje:
@davidstaheli Is anything preventing this PR to get merged?

PR #5337, opened 2019-08-22
Update download-secure-file.md

yevster:
Adding mention that files are downloaded to `$(Agent.TempDirectory)`. We have seen this information needed for customers who require uploading files in bulk or using tasks that require a directory and a file name, rather than an absolute path, such as the FTP task.


yevster:
@N-Usha, please have a look when you have a chance.

PR #5345, opened 2019-08-23
Made more clear on how to do operations using Git Credentials

brunomartinspro:
Made more clear on how to do operations using Git Credentials, generally the community doesn't read this part and starts using PATs or user:password in plain text - this made a lot of stack overflow posts..

### DIFF
--- a/docs/pipelines/yaml-schema.md
+++ b/docs/pipelines/yaml-schema.md
@@ -1637,9 +1637,9 @@ steps:
 - checkout: none
 ```
 
-# [Example](#tab/example)
-If you want to modify the current repository using Git Operations and/or load Git Submodules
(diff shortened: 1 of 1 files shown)


steved0x:
@brunomartinspro Thank you for this contribution. Due to our formatting, there can't be 2 example tabs, but you can show more than one example in a single tab. Could you move your example to go after the existing example, like this:

> [!NOTE]
> If you want to modify the current repository using Git Operations and/or load Git Submodules, make sure to give the proper permissions to "Project Collection Build Service Accounts" in the project security.

```yaml
steps:
- checkout: self
  submodules: true
  persistCredentials: true
```

brunomartinspro:
@steved0x Updated based on the recommendations, thank you 👍

PR #5357, opened 2019-08-23
Fix typo: "PATS" instead of "PATs"

jrevillas:
Just to be consistent.


chasewilson:
Great catch! Thanks for your contribution @jrevillas!

PR #5358, opened 2019-08-24
Fix omitted parameter in template docs

ruffsl:


steved0x:
@ruffsl Thank you for calling our attention to this. I am keeping this PR open for now, but we may end up replacing that whole example with something different and hopefully better. That should happen very soon, and until then I will keep this PR open in case we decide to go with this in the interim. Thank you for your contribution.

steved0x:
@ruffsl I am closing this one, we still have an item on the backlog and we are going to replace that section. Thank you for your contribution and feedback.

PR #5364, opened 2019-08-26
fix path to version-selector img

Styxxy:


chasewilson:
@Styxxy, Thanks again for the help :) I've actually got a PR to fix this in right now. I'll be closing this one but you should see the change shortly!

PR #5365, opened 2019-08-26
remove weird/non-latin characters

Styxxy:


chasewilson:
@Styxxy, Thank you for the contribution!

PR #5439, opened 2019-08-31
fix: MD046/code-block-style

nschonni:
Ensure consistent fended code block style.
The implicit indent fences sometimes accidentally swallow content not intended as code blocks.

### DIFF
--- a/docs/pipelines/apps/_shared/create-release-azure-webapp-container.md
+++ b/docs/pipelines/apps/_shared/create-release-azure-webapp-container.md
@@ -39,27 +39,19 @@ ms.topic: include
    - **Startup command**: Start up command for the container.
 
    <!--
-
    1. Open the **Tasks** tab and select the **Stage 1** item.
    Configure the linked settings as follows:
-
    - **Azure Subscription**: Select a connection from the list under **Available Azure Service Connections** or create a more restricted permissions connection to your Azure subscription.
      If you are using Azure Pipelines and if you see an **Authorize** button next to the input, click on it to authorize Azure Pipelines to connect to your Azure subscription. If you are using TFS or if you do not see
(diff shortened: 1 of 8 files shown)


steved0x:
@nschonni Thank you for helping us to make our docs better. In the past I was able to approve larger PRs like this, but now instead of having a central person review PRs, we are moving to a model where the owners of the different services review the PRs in their areas. Big PRs like this are tougher to review. I believe that you are using some automated tools to help sweep the docs and make changes, is it possible to break them into multiple runs that are based on the folders under docs? For example, one for Boards, one for Repos, one for Pipeline, and like that? This would allow each service owner to review and approve the ones in their area, without depending on other area owners. I'll see if I can get the contributor guide updated to include this guidance as well.

I know this is a hassle (and I am hoping that if you are using an automated mechanism for making these changes, it can accommodate this new workflow). Thank you.

nschonni:
I'm not using any automated mechanism, but I can try and split the commit by area

nschonni:
@steved0x as a start, I split the commits so that each one is for one of the areas in the CODEOWNERS. 
I also removed the .markdownlint.json change because that can't happen till all the changes land. Let me know if you want 11 PRs, or if there is a logical grouping

PS: noticed that there is no CODEOWNER for `docs/reference/` or `docs/extend/`

nschonni:
I think the `**` also need to be removed from the paths so the CODEOWNSERS file works. EX: https://github.com/Azure/azure-rest-api-specs/blob/master/CODEOWNERS

steved0x:
@nschonni 11 PRs is better for us :) The best logical grouping is by root path like `repos`, `pipelines`, etc... OK Looking at your 11 commits, those are **perfect**, if they were 11 PRs.

nschonni:
OK, it's broken up. CodeOwners isn't working. I think the prbot used on other repos that assigns the authors seems more reliable

PR #5538, opened 2019-09-10
fix: MD047/single-trailing-newline release-notes/

nschonni:
Files should end with a single newline character


nschonni:
Did this by running `eclint fix -n release-notes/`

Issue #5505, opened 2019-09-05
DownloadPipelineArtifact@2 - YAML not consistent with parameter table


tillig:
The YAML snippet shows the following inputs for `DownloadPipelineArtifact@2`:

- `buildType`
- `project`
- `definition`
- `specificBuildWithTriggering`
- `buildVersionToDownload`
- `branchName`
- `pipelineId`
- `tags`
- `artifactName`
- `itemPattern`
- `targetPath`

However, the table below it lists:

- `source`
- `preferTriggeringPipeline`
- `artifact`
- `patterns`
- `path`
- `project`
- `pipeline`
- `runVersion`
- `runBranch`
- `runId`

It's like there was a copy/paste error or something.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: b03d67d0-8738-72e6-dbe7-324c1784d4fc
* Version Independent ID: bbb4d774-cd98-3ecb-7ccf-5f49e6f2b425
* Content: [Download Pipeline Artifact task - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact?view=azure-devops#feedback)
* Content Source: [docs/pipelines/tasks/utility/download-pipeline-artifact.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/utility/download-pipeline-artifact.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @willsmythe
* Microsoft Alias: **wismyth**

tillig:
`tags` is also described as "coma-delimited" which is either a typo or a really, really long time between tags. :)

ShadowXVII:
Looks like it's been updated now. The old terms have become aliases in the table.

cxwtool:
This issue hasn't been updated in more than 180 days, so we've closed it. If you feel the issue is still relevant and needs fixed, please reopen it and we'll take another look. We appreciate your feedback and apologize for any inconvenience.

PR #5639, opened 2019-09-17
DownloadPipelineArtifactV2 - Adding aliases for parameters

bchurchill:
See #5505.  The underlying issue is that there are multiple names for several of the parameters, and these different names are confusing (e.g. there's a pipeline parameter that takes an id, and also pipelineId parameter, but these do different things!).  In addition to the work that's already been done to clarify the documentation, this PR adds the names of aliases for all of the arguments.  This will especially help those who already have code written understand what the available parameters are.

### DIFF
--- a/docs/pipelines/tasks/utility/download-pipeline-artifact.md
+++ b/docs/pipelines/tasks/utility/download-pipeline-artifact.md
@@ -28,19 +28,19 @@ Use this task in a build or release pipeline to download pipeline artifacts from
 ## Arguments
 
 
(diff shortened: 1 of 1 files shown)


PRMerger9:
@bchurchill : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
@bchurchill Thank you for submitting the PR. The arguments that are listed and updated in the pull request are no longer in the article in master. We'll close the PR because it can't be merged as-is.

PR #5661, opened 2019-09-19
Add description about Microsoft-hosted agent.

rikat-ms:
We can use Microsoft-hosted agent to deploy applications, but sometimes there are firewall/proxy/restricted IP configuration between DevOps and destinations. In this case, users must know VMs for Microsoft-hosted agent use any of Azure datacenters IP range. This has impact for users and they may need to consider using Self-hosted agent. So this page should have the link to Microsoft-hosted agent page so that users can find the information easily.


chasewilson:
@rikat-ms thanks for your help in this! Merging this in

PR #5679, opened 2019-09-19
Remove duplicate entry in field types

zeusk:
See #5678


chasewilson:
@zeusk Thanks for the input! :)

PR #5691, opened 2019-09-20
Adding multiple snapshot version handling.

yevster:
When using Azure Artifacts as a source, it is possible to get multiple binary files for a single version. This can break downstream build, as happened with a customer we are working with. I'm adding a note for how to address.

### DIFF
--- a/docs/pipelines/release/artifacts.md
+++ b/docs/pipelines/release/artifacts.md
@@ -339,8 +339,9 @@ The following features are available when using Azure Artifacts sources:
 |   Artifact download    | By default, packages are downloaded to the agent. You can configure an option in the stage to [skip the download](../process/phases.md#agent-phase) of artifacts. |
 
 <p />
+<h4 id="mavensnapshots">Handling Maven Snapshots</h4>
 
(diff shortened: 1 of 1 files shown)


docs/pipelines/release/artifacts.md
```diff
@@ -340,6 +340,12 @@ The following features are available when using Azure Artifacts sources:
 
 <p />
 
+*Note:* when obtaining Maven artifacts and the artifact is a snapshot build, mulitple versions of that snapshot may be downloaded at once (e.g. `myApplication-2.1.0.BUILD-20190920.220048-3.jar`, `myApplication-2.1.0.BUILD-20190820.221046-2.jar`, `myApplication-2.1.0.BUILD-20190820.220331-1.jar`). You will likely need to add additional automation to keep only the latest artifact prior to subsequent deployment steps. This can be accomplished with the following PowerShell snippet:
```
chasewilson: This looks great and thank you for the input! The only thing I would say is note to use the Note, since you have a PowerShell snippet to copy I would let it be a section on it's own. 
However, if you feel that a 'Note' here is better, use the format:
> [!NOTE]
> <Your note string here.>

yevster: Replaced with a subheading.

PR #5693, opened 2019-09-21
Update build-image.md

upendra25312:
Note: If you are using .net core then replace the  imageName: 'pipelines-javascript-docker 'with  imageName: 'pipelines-.net-core-docker'


shashankbarsin:
@upendra25312 the sample mentioned in the documentation is just to showcase how image build can happen for any repository containing a Dockerfile. We don't want to exhaustively enumerate every language and provide a sample for that language in the documentation.

PR #5767, opened 2019-09-27
Changing text to reflect the use of Virtual Support agent for region …

ramakoni1:
…changes

Updated info on regions changes based on a thread with supportability team.


ramakoni1:
#sign-off

PR #5813, opened 2019-10-02
Update advanced-code-search-syntax.md

sunar91:
Adding changes for filename search for files without extension


sunar91:
Search without file extension changes

Issue #5723, opened 2019-09-24
"Test on the Android emulator" section is outdated


sbolotovms:
Hi,

this section here: https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/android?view=azure-devops#test-on-the-android-emulator
is outdated and works for previous hosted VM generation - macOS 10.13. Since then, Azure provides VM macOS 10.14 definition, for which this no longer applies. Thanks!

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 454383d9-78ce-6203-2926-3449842d2f9f
* Version Independent ID: 947f872e-95b9-1691-acf0-e1a3f73b1e75
* Content: [Build, test, and deploy Android apps - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/android?view=azure-devops#test-on-the-android-emulator)
* Content Source: [docs/pipelines/ecosystems/android.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/ecosystems/android.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @thomps23
* Microsoft Alias: **dastahel**

sphanley:
I'm part of a team who's trying to run Android tests on Azure DevOps - is there any info available as to what aspects of these instructions are no longer applicable for the 10.14 VM agents, and what the new recommended steps look like?

cxwtool:
This issue hasn't been updated in more than 180 days, so we've closed it. If you feel the issue is still relevant and needs fixed, please reopen it and we'll take another look. We appreciate your feedback and apologize for any inconvenience.

PR #5881, opened 2019-10-08
Uncomment incorrectly commented-out line in code sample to launch emulator

sphanley:
In the code snippet describing the code required to launch an Android Emulator on the Azure Pipeline agent, the line which invokes `sdkmanager` to download the emulator image is commented out. However, this emulator image is not preloaded on the agent VMs, so this line must be uncommented for these instructions to work. In the solution to [this Developer Community post](https://developercommunity.visualstudio.com/content/problem/340547/azure-devops-unable-to-run-the-android-emulator-on.html), similar code is provided by a Microsoft rep with the line in question uncommented.

### DIFF
--- a/docs/pipelines/ecosystems/android.md
+++ b/docs/pipelines/ecosystems/android.md
@@ -121,7 +121,7 @@ Don't forget to arrange the emulator parameters to fit your testing environment.
 #!/usr/bin/env bash
 
 # Install AVD files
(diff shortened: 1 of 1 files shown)


sphanley:
It's worth noting that #5723 suggests that this entire codeblock is outdated on the MacOS 10.14 agents. However, with the modification suggested in this PR, it's been my experience that this code block still works as expected on the 10.14 agents, and I'm not aware of any newer mechanism for launching an emulator on the pipeline agents.

juliakm:
@sphanley Thanks for your contribution!

PR #5914, opened 2019-10-09
Add secret masking info

damccorm:
@vtbassmatt could you take a look?

### DIFF
--- a/docs/pipelines/process/_shared/set-secrets.md
+++ b/docs/pipelines/process/_shared/set-secrets.md
@@ -20,6 +20,8 @@ Secrets are available on the agent for tasks and scripts to use **(so be careful
 > 
 > We will not ever mask substrings of secrets. If, for example, "abc123" is set as a secret, "abc" will not be masked from the logs.
 > This is to avoid masking secrets at too granular of a level, making the logs unreadable.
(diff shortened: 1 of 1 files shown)


docs/pipelines/process/_shared/set-secrets.md
```diff
@@ -17,6 +17,9 @@ Secrets are available on the agent for tasks and scripts to use **(so be careful
 > We make an effort to mask secrets from appearing in Azure Pipelines output, but it's not foolproof. **Never echo secrets as output.**
 > Some operating systems log command line arguments. **Never pass secrets on the command line.**
 > Instead, we suggest that you map your secrets into environment variables.
+> 
+> We will not ever mask substrings of secrets. If, for example, "abc123" is set as a secret, "abc" will not be masked from the logs.
```
zachariahcox: we might need to add: 

For this reason, **secrets should not contain structured data**.
If the secret were the string ```{ "user": "zac", "password": "secrets" }```,  we would not mask the substring "secrets" when it appears by itself in the logs.

damccorm: Good idea, updated appropriately.

Issue #5901, opened 2019-10-09
Missing documentation for apksignerFile


ChristianKleineidam:
This file misses documenting what the apksignerFile option does.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 4b1fbcd8-d799-879e-d672-9dee164401ef
* Version Independent ID: 7208d03d-9042-7753-ab4d-d1ad8e99ef1d
* Content: [Android signing build and release task - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/android-signing?view=azure-devops)
* Content Source: [docs/pipelines/tasks/build/android-signing.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/build/android-signing.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @DavidStaheli
* Microsoft Alias: **dastahel**

PR #5967, opened 2019-10-14
Update android-signing.md with latest verbiage

sphanley:
This documentation is currently missing the verbiage for apksignerFile. I also included minor verbiage changes from the latest task version, and revised the row labels in the table to match the actual parameters, to be more consistent with other docs.

Fixes https://github.com/MicrosoftDocs/vsts-docs/issues/5901

### DIFF
--- a/docs/pipelines/tasks/_shared/yaml/AndroidSigningV3.md
+++ b/docs/pipelines/tasks/_shared/yaml/AndroidSigningV3.md
@@ -9,7 +9,7 @@
     #apksignerKeystorePassword: # Optional
     #apksignerKeystoreAlias: # Optional
     #apksignerKeyPassword: # Optional
-    #apksignerArguments: '--verbose' # Optional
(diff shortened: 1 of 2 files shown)


sphanley:
This should resolve #5901.

sphanley:
Closing this because I'm suspicious that the defaults for apksignerArguments shouldn't actually be changed to `-verbose -sigalg MD5withRSA -digestalg SHA1`. That's what's listed in the info tooltip on the task, but per apksigner, those don't seem to be valid args, and the actual prefilled value is `--verbose`. I suspect it's the tooltip that's incorrect, not the documentation. I'm going to revise to simply add the missing info for `apksignerFile`.

sphanley:
Confirmed: `-verbose -sigalg MD5withRSA -digestalg SHA1` was the default `apksignerArguments` value for the 2.* task, but the 3.* task version defaults to `--verbose`. Reverted the incorrect changes. Based on this, the 3.* task itself should be updated so that the tooltip accurately lists the current defaults.

Issue #6053, opened 2019-10-18
"configuration" does not work


jeremy-jameson:
It appears the **configuration** key in the **inputs** list is obsolete.

For example, if you specify the following in azure-pipelines.yml, Azure DevOps will actually build the default configuration (i.e. **Debug**) for your projects (not the **Release** version):
```
variables:
  buildConfiguration: "Release"

steps:
  - task: DotNetCoreCLI@2
    displayName: "Build"
    inputs:
      command: build
      configuration: $(buildConfiguration)
```

To achieve the desired result, you must specify the following:
```
variables:
  buildConfiguration: "Release"

steps:
  - task: DotNetCoreCLI@2
    displayName: "Build"
    inputs:
      command: build
      arguments: "--configuration $(buildConfiguration)"
```
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 1d1d9884-19f7-592b-f47c-8cf100639ff0
* Version Independent ID: 555c3826-60d9-2589-54af-850efbfbb13a
* Content: [.NET Core CLI task - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops#feedback)
* Content Source: [docs/pipelines/tasks/build/dotnet-core-cli.md](https://github.com/MicrosoftDocs/vsts-docs/blob/master/docs/pipelines/tasks/build/dotnet-core-cli.md)
* Product: **devops**
* Technology: **devops-cicd**
* GitHub Login: @pulkitaggarwl
* Microsoft Alias: **puagarw**

msebolt:
@jeremy-jameson Thank you for the feedback, assigning this to the author for review.

cpdohert:
This doesn't work for dotnet pack.  Given this:

```
- task: DotNetCoreCLI@2
            displayName: "Build and Pack the Library"
            inputs:
              command: "pack"
              arguments: "--configuration Release -p:PackageVersion=$(GitVersion.NuGetVersion) -p:Version=$(GitVersion.AssemblySemVer)"
              versioningScheme: byEnvVar
              versionEnvVar: GitVersion.NuGetVersion
              packagesToPack: ${{ parameters.ProjectPath }}
```

The task completely ignores the `arguments:` option entirely and constructs the command line thusly:
```
##[debug]arguments:
##[debug]   pack
[...]

PR #6054, opened 2019-10-18
Remove "configuration" key from "inputs" list (#6053)

jeremy-jameson:
My editor is configured to remove trailing spaces from lines. When I diff'ed my changes in Visual Studio Code, it only showed one change (i.e. the deleted line). I suppose I could revert my change and modify in Notepad, but I'm not going to bother.


ktoliver:
Hi @davidstaheli - Could you review the proposed changes? Thank you.

PR #6137, opened 2019-10-25
Document pwsh input of PowerShell@2

chwarr:


chwarr:
I based this on reading the other documentation. I haven't experimented or checked the code to validate this.

PR #6310, opened 2019-11-08
Added NOTE about enabling SQL Server Auth mode

gregschunemann:
I've made this mistake a couple times now (one too many).  Without enabling SQL logins on the server, we receive the following error:
image
After checking the inbound port rules, and finding that this port is open to inbound traffic I was stumped.  I think that this note will be useful for people that are less familiar with SQL server such as myself.


PRMerger13:
@gregschunemann : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6312, opened 2019-11-08
Corrections for work item operations permissioning

KimPlausible:
PCA or TFA permissions will not take precedence over others for work item operations, such as deletion. Any Deny inherited from group membership will trump these Admin rights.

Edited all references.


PRMerger14:
@KimPlausible : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6313, opened 2019-11-08
Limitation on state-based rules

KimPlausible:
Added note to explain limitation on state-based rules. We don't support more than 1 condition for state-based rules.


PRMerger15:
@KimPlausible : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6345, opened 2019-11-12
Fix typos in deployment-jobs.md (pre-deploy --> preDeploy)

gkalpak:
Using `pre-deploy` (as shown in the examples) causes the pipeline to fail with:
`Unexpected value 'pre-deploy'`


PRMerger19:
@gkalpak : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6414, opened 2019-11-19
Clarify that variable groups can be used for passing secrets into YAML pipelines

johndowns:
This PR clarifies a point of confusion that I have seen with several customers, which is how to pass secrets into multi-stage YAML pipelines that have release stages.

A common use case is to use multi-stage YAML pipelines to deploy application to multiple environments, and frequently applications need to have environment-specific secrets flowed into their configuration systems through a pipeline (e.g. an API key for an external API). Currently, it appears that variable groups are the best and recommended approach to achieve this, so clarifying this within the documentation would be advantageous.


PRMerger14:
@johndowns : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6539, opened 2019-12-03
Update Section on ConnectALL

igoose:
Updated info on ConnectALL. Would like to update logo as well


PRMerger6:
@igoose : Thanks for your contribution! The author(s) have been notified to review your proposed change.

KathrynEE:
@igoose  - If you open an issue for this topic and append the logo you want used, I'll update the topic to use that logo.

PR #6546, opened 2019-12-03
Update install-extensions.md

msdnguy:
Users can no longer be assigned to extensions.  This change is related to the fact that we no longer charge for 3rd party extensions and Test and Package Management extensions no longer exist.


PRMerger12:
@msdnguy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6571, opened 2019-12-05
wrong parameter names

tfenster:
Parameters `publisher-name` and `extension-name` don't work, instead `publisher-id` and `extension-id` seem to be correct


PRMerger17:
@tfenster : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley:
#sign-off

PR #6711, opened 2019-12-17
Fix double-encoded ampersands

patridge:
I could be misunderstanding what is being conveyed here, but I don't think these `>` symbols should be encoded like this in a `<code>` block.
Screenshot of docs showing several date queries with HTML-encoded ampersands


PRMerger9:
@patridge : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6770, opened 2019-12-24
replacing sample queries URL account path with organization

JBrLloyd:
some of the Sample Query URL paths contains an old name {account} and should be updated to use {organization} to be consistent with other MS Docs pages


PRMerger20:
@JBrLloyd : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #6883, opened 2020-01-10
Fix isParameterizedJob parameter name

Ekhozie:
Documentation for parameter isParameterizedJob do not match the name in the example (and `ParameterizedJob` does not work in a pipeline, we need to use the name `isParameterizedJob`)


PRMerger13:
@Ekhozie : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7021, opened 2020-01-22
List of branches to trigger pipelines doesn't work

piranna:
Documentation give the impression that you can use a plain list of branches to define the ones that will trigger the pipeline, but that doesn't works. It needs to use the "long" version, specifying the `branches` and `includes` nodes. This is similar to how it's defined the default value some lines under.

### DIFF
--- a/docs/pipelines/repos/azure-repos-git.md
+++ b/docs/pipelines/repos/azure-repos-git.md
@@ -118,8 +118,10 @@ You can control which branches get CI triggers with a simple syntax:
 
 ```yaml
 trigger:
-- master
-- releases/*
+  branches:
+    include:
+    - master
+    - releases/*
 ```
 
 You can specify the full name of the branch (for example, `master`) or a wildcard (for example, `releases/*`).


PRMerger13:
@piranna : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x:
@piranna Thank you for this contribution - let me review these and we'll get it merged. Thank you.

piranna:
You are welcome! :-)

steved0x:
@piranna Can you share your trigger section that isn't working, I just made a pipeline with the following triggers, where the repo is an Azure Repos Git repo.

```
trigger:
- master
- users/jamalh
```

and was able to trigger a build by making a push to `users/jamalh` (in fact I added that line in the newly created `users/jamalh` branch and pushed it.) 

Do you have any other conditions in your trigger? I believe the shortcut syntax only works in you are simply listing branches. If so I can try and clarify that in the documentation.

steved0x:
@piranna I am closing this PR, but if you still have trouble with using the shortcut branch syntax, please reopen it or file an issue by clicking one of the options on the **Feedback** section at the bottom of our articles, and we can take another look. Thank you.

image

PR #7073, opened 2020-01-27
Update yaml-schema.md

droessmj:
Specify the cron schedule is executed in UTC time

### DIFF
--- a/docs/pipelines/yaml-schema.md
+++ b/docs/pipelines/yaml-schema.md
@@ -1402,7 +1402,7 @@ Learn more about [scheduled triggers](build/triggers.md?tabs=yaml#scheduled-trig
 
 ```yaml
 schedules:
-- cron: string # cron syntax defining a schedule
+- cron: string # cron syntax defining a schedule in UTC time
   displayName: string # friendly name given to a specific schedule
   branches:
     include: [ string ] # which branches the schedule applies to


PRMerger14:
@droessmj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

vtbassmatt:
This isn't correct, it's scheduled in the time set by the organization.

droessmj:
Appreciate the feedback, but that statement directly conflicts with the behavior I experienced when working with this yesterday. 

Our organization runs with UTC-6, 
image, but I needed to set a schedule for 18:20 yesterday for it to trigger at 12:20PM CT. 

image

Let me know if you'd prefer I open an issue, but this PR was submitted in an attempt to reduce future occurrences. 

Only other potential impact I can come up with offhand is that the pool in use is a Windows Hosted Agent rather than a corporately hosted pool. Not sure if that would make a difference or not on scheduling?

steved0x:
@droessmj It is UTC time, that is clarified [here](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml#scheduled-triggers) but I think it is a good idea to update the comment like you did so I'm approving this PR.

steved0x:
#sign-off

PR #7132, opened 2020-01-31
Consolidate exceptions to regions

lupine86:
Mac hosted builds are run in either the United State or data centers in Europe.

This change attempts to call out all the exceptions to single region in one area.


PRMerger20:
@lupine86 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley:
#sign-off

PR #7136, opened 2020-01-31
Remove extra WIQL code

patridge:
I could be wrong, but it appears the WIQL code I removed was a typo. The line had mismatched casing from the rest of the query and appears twice.


PRMerger8:
@patridge : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7331, opened 2020-02-18
Update resolve-orphaned-organization.md

pazand:
Change 1: Since multiple types of admins are being referenced in this article (PCA, Owner, Azure Global Admin, Privileged Role Admin, Azure DevOps Admin) i think it's better to be clear here 

Change 2:  The Azure DevOps Admin doesn't actually give you the ability to change the org owner in the previously linked doc. Unless the stated Azure DevOps Admin is also a PCA/Org owner, the only way they can take over the org is by following the documentation on this page.


PRMerger9:
@pazand : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley:
#sign-off

PR #7334, opened 2020-02-18
Added note on copy/paste

ChristopherMank:
Added note that you cannot copy/paste from a previous comment. While they look the same, it's not a true mention and doesn't send an email.


PRMerger20:

chcomley:
#sign-off

PR #7484, opened 2020-03-04
update java-function.md

seyedk:
line 95 should not have the /$(appName) in the path, as the  path to publish in previous task does not include it, otherwise, it caues the pipeline to fail.


PRMerger17:
@seyedk : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x:
#sign-off

PR #7557, opened 2020-03-10
Clarify nuget.exe usage

zivkan:
The YAML comment says that the task uses nuget.exe and should be used with .NET Framework projects (although technically this is incorrect, it should be used with non-SDK style projects. SDK style projects can target the .NET Framework, in which case the dotnet cli or msbuild should be used, not nuget.exe. For packing projects, even non-SDK style projects using PackageReference should use msbuild and not nuget.exe). Anyway, that very useful snippet of information is not visible when viewing the page because it's in a code block that is too wide for the page and therefore is hidden unless you scroll to the right.


PRMerger15:
@zivkan : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x:
#sign-off

PR #7736, opened 2020-03-28
escape example URLs

meganbradley:
CATS used an allow list to exclude example links and other non-clickable links from 404 reports. The new broken link report doesn't support an allow list, so this PR is to escape example links as inline code.


PRMerger20:
@meganbradley : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #7873, opened 2020-04-09
Explain global/job level variable mapping for secrets

EugenMayer:
It is not clear in the docs, nor the example, nor is it easy to debug, if `global` or `job` level mapped variables are available as ENV variables in the steps/tasks - they are not.

It is important to notice, since if using as a template variable like explained in the inputs of the copy task, it will work.


PRMerger7:
@EugenMayer : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
#sign-off

PR #8115, opened 2020-04-28
Corrected runOnce strategy example

rudfoss:
Corrected output variables schema for runOnce strategy and added an example.

It previously indicated that one should use the `lifecycle-hookname` to access output variables in another job. This is in fact false and the example from [this document](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops&tabs=yaml%2Cbatch#set-a-multi-job-output-variable) shows the correct implementation.

Ref: https://stackoverflow.com/questions/61476512/azure-devops-pipeline-define-variable-in-deployment-and-reuse-in-subsequent-job/61477357#61477357

### DIFF
--- a/docs/pipelines/process/deployment-jobs.md
+++ b/docs/pipelines/process/deployment-jobs.md
@@ -354,12 +354,12 @@ Define output variables in a deployment job's [lifecycle hooks](https://docs.mic
 
 While executing deployment strategies, you can access output variables across jobs using the following syntax.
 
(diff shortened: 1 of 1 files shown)


PRMerger19:
@rudfoss : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
Thanks @rudfoss! 

#sign-off

PRMerger14:

PR #8142, opened 2020-04-29
Mention that Maven task overwrites MAVEN_OPTS

winniehell:
For the Maven task, the default for `mavenOpts` / `mavenOptions` is not to fall back to the environment variable but a fixed value of `-Xmx1024m`. Therefore the caching configuration does not work out of the box with the Maven task. This updates the documentation with an example for a workaround.

### DIFF
--- a/docs/pipelines/release/caching.md
+++ b/docs/pipelines/release/caching.md
@@ -283,6 +283,15 @@ steps:
 - script: mvn install -B -e
 ```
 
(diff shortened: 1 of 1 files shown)


PRMerger10:
@winniehell : Thanks for your contribution! The author(s) have been notified to review your proposed change.

winniehell:
@ramiMSFT @ktoliver Thanks for the amazingly fast response time! :rocket: :heart:

PR #8238, opened 2020-05-07
clarify auto-complete functionality

evandigby:
I would not expect auto-complete to require "optional" policies by default. I propose clarifying the documentation in this regard.

### DIFF
--- a/docs/repos/git/branch-policies.md
+++ b/docs/repos/git/branch-policies.md
@@ -216,7 +216,7 @@ No. After you set up a required branch policy, you can't directly push changes t
 
 #### What is auto-complete?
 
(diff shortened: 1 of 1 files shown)


PRMerger12:
@evandigby : Thanks for your contribution! The author(s) have been notified to review your proposed change.

vtbassmatt:
This isn't true, or at least isn't globally true. We have optional policies in our own repo and autocomplete doesn't wait on them. In fact, some can even fail and autocomplete still works.

@evandigby are you still experiencing this? There might be a setting I'm unaware of (new to the policies area, despite a long history with the Git server).

ktoliver:
Hi @vtbassmatt Do you want to keep this pull request open? Thanks.

vtbassmatt:
@ktoliver it can be closed.

PR #8442, opened 2020-05-25
Modify publishWebProjects description.

angusho321:
Hello, I think the description of "publishWebProjects" should be amended to "If true, the task will only try to find ...". Because when I use this task, although my project is a web project, I also need to publish other projects.


PRMerger16:
@angusho321 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mijacobs:
@steved0x @ramimsft @juliakm  -- could you take a look at this PR?

steved0x:
@angusho321 There is a conflicting PR that has different wording to resolve your issue, here is a link to the updated section: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/build/dotnet-core-cli?view=azure-devops and here is the updated property:

> If `true`, the `projects` property value will be skipped and the task will try to find the web projects in the repository and run the publish command on them. Web projects are identified by presence of either a web.config file or wwwroot folder in the directory. Note that this argument defaults to `true` if not specified.

If this newer updated description is not correct, please let us know by choosing "This page" from the feedback on that article.

image


Thank you for your contribution and feedback.

PR #8587, opened 2020-06-09
Mention pipeline artifact tasks clearly in docs

EliiseS:
When looking into what are the options for getting artifacts from other pipelines in templates, it wasn't very clear that you could use the tasks instead of the resources too. 

I'm still unsure about what the difference between using resources vs tasks.


PRMerger6:
@EliiseS : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
Thanks @EliiseS!

#sign-off

PRMerger16:

PR #8666, opened 2020-06-18
Update allow-list-ip-url.md for Graph endpoint

bpcastilho:
graph.windows.net will be deprecated and we need to start using graph.microsoft.com endpoint instead.


chcomley:
#sign-off

PR #8681, opened 2020-06-19
Update condition docs to cover cancelation

mjroghelia:
Call out that the succeeded function also
returns false when the pipeline is canceled.
Update the example using a variable in a condition
to reflect the best practice of combining it with
the succeeded function. The example as previously
written encouraged a pattern that leads to stages/jobs
executing even when the pipeline is canceled, which
is not usually what people want.

### DIFF
--- a/docs/pipelines/process/conditions.md
+++ b/docs/pipelines/process/conditions.md
@@ -55,7 +55,7 @@ stages:
       - script: echo Hello Stage A!
 
 - stage: B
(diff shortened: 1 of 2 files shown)


PRMerger6:
@mjroghelia : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
Thanks @mjroghelia! This is helpful to know!

#sign-off

PR #8742, opened 2020-06-30
Fix PowerShell commands for newer modules

jgallucci32:
Newer versions of PowerShell do not support the `name -eq $projectName` parameter as well as the `@> > >` usage for setting the Content-Type in the REST header. This updates the PowerShell examples to work with modern versions.


PRMerger17:
@jgallucci32 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #8752, opened 2020-07-01
Wrap GitHttpClient code samples in using(...)

mcthib:
The documentation over at https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2013/dn413500(v=vs.120) is from 2014 and no longer updated. The base class VssHttpClientBase now implements IDisposable, so it's important to demonstrate proper usage of this object.


PRMerger17:
@mcthib : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #8925, opened 2020-07-24
Update configure-test-results-trend.md

alsulli:
Under "prerequisites" I added that the Test Plans service needs to be enabled in order for this widget to work. If it isn't enabled, then the widget will throw an error "Access not supported".

This should be in the prereqs for this widget, as we state that pipelines service must be enabled, so we should also state test plans must be enabled.


PRMerger17:
@alsulli : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #8955, opened 2020-07-29
Fix YAML in service connection webhook example

theBNT:
Since the comment is in a multi-line inline code section, its not parsed correctly and throws the following error when trying the example 1:1:

`Encountered error(s) while parsing pipeline YAML:
/azure-pipelines.yml (Line: 14, Col: 13): Unexpected symbol: '<WebhookAlias>'. Located at position 12 within expression: parameters.<WebhookAlias>.<JSONPath>. For more help, refer to https://go.microsoft.com/fwlink/?linkid=842996`

Solution: move comment above section.

I did not find a way to make inline comments (e.g. powershell in this case) to work, so this was the best i could come up with.

### DIFF
--- a/release-notes/2020/includes/pipelines/sprint-172-update.md
+++ b/release-notes/2020/includes/pipelines/sprint-172-update.md
@@ -62,8 +62,9 @@ steps:
 - task: PowerShell@2
   inputs:
     targetType: 'inline'
+    ### JSON payload data is available in the form of ${{ parameters.<WebhookAlias>.<JSONPath>}}
     script: |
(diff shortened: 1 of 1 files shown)


PRMerger18:
@theBNT : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Sid-ah:
@ashokirla can you please take a look?

vtbassmatt:
#sign-off

PRMerger12:
@vtbassmatt: I'm sorry - only the author of this article, @sid-ah, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #8970, opened 2020-07-31
Fix: Authentication schema's examples are not correct

johnlokerse:
I had some issues using the examples (mainly the Token based). The service connection wouldn't show up in my Service Connections list. After trying out some solutions I found the fix. 

Under the "properties" it is missing the "authenticationSchemes" and  the"type" property. After adding this, it worked properly. Also, I updated the "type" and "targets" instead of referring to a schema it now refers to a type.

Also see this documentation about it: [GitHub - Azure Pipelines Service Endpoints](https://github.com/microsoft/azure-pipelines-extensions/blob/master/docs/authoring/endpoints/serviceEndpoints.md)

### DIFF
--- a/docs/extend/develop/auth-schemes.md
+++ b/docs/extend/develop/auth-schemes.md
@@ -33,43 +33,48 @@ Default authentication header used is: "Basic {{ #base64 endpoint.username \":\"
 {
     "id": "endpoint-auth-scheme-basic",
     "description": "Basic Authentication based endpoint authentication scheme",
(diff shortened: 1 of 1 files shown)


PRMerger19:
@johnlokerse : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger6:
@johnlokerse : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley:
#sign-off

PR #9056, opened 2020-08-12
document /format:xml parameter

FlorisDevreese:
This parameter was not documented at all. It can be important to retrieve important information about the workspaces that was not available in the other `/format` parameter options.


PRMerger14:
@FlorisDevreese : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #9169, opened 2020-08-27
Fix yaml example rendering

radujipa:
The YAML syntax colouring seems to be getting confused after the first `script` field on this page https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions?view=azure-devops&tabs=yaml#ive-got-a-conditional-step-that-runs-even-when-a-job-is-canceled-how-do-i-manage-to-cancel-all-jobs-at-once

I suspect this may be due to the concat operator `>`, which is valid YAML. I've inlined the oneliner hoping this would solve the issue.


PRMerger8:
@radujipa : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
Thanks @radujipa! 

#sign-off

juliakm:
#sign-off

PR #9197, opened 2020-09-01
Update run-number.md

farrrb:
See https://developercommunity.visualstudio.com/content/problem/303198/the-buildid-variable-does-not-get-replaced-anymore.html

### DIFF
--- a/docs/pipelines/process/run-number.md
+++ b/docs/pipelines/process/run-number.md
@@ -74,7 +74,7 @@ The following table shows how each token is resolved based on the previous examp
 | Token | Example replacement value |
 | ----- | ------------------------- |
 | `$(Build.DefinitionName)` | CIBuild<br /><br />Note: The pipeline name must not contain invalid or whitespace characters.|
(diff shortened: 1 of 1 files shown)


PRMerger12:
@farrrb : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MarkIannucci:
@farrrb , I think the problem you may have run into was that you were trying to use a build number token (in your case BuildID) outside of the run-number definition.  That doesn't seem to be allowed.  I ran into this challenge independently of you last night and submitted #9214 to try to improve the documentation. 

In the rest of my script I used the variable I used the `$(Build.BuildNumber)` variable to get a CI created SemVer in my nuget packages.  In that process I also found an [issue in the NugetCommand@2](https://github.com/microsoft/azure-pipelines-tasks/issues/13509), which may be an issue for you as well.

juliakm:
Thanks @MarkIannucci. I agree that this is likely what's going on. Here's an example with a working `$(BuildID)` for reference. 

```yaml
name: $(TeamProject)_$(BuildID)_$(SourceBranchName)_$(Date:yyyyMMdd)$(Rev:.r)

steps:
  - script: echo $(Build.BuildNumber) # outputs customized build number like project_def_master_20200828.1

```

PR #9237, opened 2020-09-08
Fix: Location of Connect to feed action

pruthvip:
Fix: Location of Connect to feed action which is on Artifacts Page and not on the Packages page.


PRMerger6:
@pruthvip : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #9481, opened 2020-10-14
Fix copy typo in docs/pipelines/index.yml 

greg-bennett:
fixed typo


ktoliver:
@greg-bennett Thanks for the typo fix.

PR #9535, opened 2020-10-23
Fix wrong code sample in PAT

kissge:
The original code gives wrong encoded value because it has '\n' at last and Azure DevOps server doesn't treat trailing whitespace characters.
By replacing `echo` with `printf`, the string would contain no '\n' at last, fixing the issue.


PRMerger15:
@kissge : Thanks for your contribution! The author(s) have been notified to review your proposed change.

vtbassmatt (approved):

PR #9576, opened 2020-10-29
Add detail around Collection Admin

ChristopherMank:
Add detail that a user doesn't need to be a collection admin to perform this action.


PRMerger10:

PR #9577, opened 2020-10-29
Add detail around collection admins

ChristopherMank:
Add detail that you don't need to be a collection admin in Azure DevOps to perform these actions.


PRMerger12:

PR #9584, opened 2020-10-31
Make list of agentless tasks visible

tfenster:
The list of agentless tasks only appears if you select the "Classic" tab. As the list is relevant for YAML as well, it should appear outside of the tabbed area


PRMerger17:
@tfenster : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
#sign-off

PR #9663, opened 2020-11-12
Update scheduled-triggers.md

Lampei:
I'm adding this "Tip" to the documentation as one of our devs ran into this issue when using days of the week.  The UI has checkboxes for Mon - Sun and the example above has Mon-Fri.  Following this logic, we used Mon-Sun in the cron task itself, but this then creates a schedule of only two days, Sun and Mon.  This did not seem, per se, to be an actual bug with the code itself, but more of a need to understand how it was generated.
Using `0 5 * * Mon-Sun` generates scheduled runs of midnight EST on Sun and Mon.
Using `0 5 * * Sun-Sat` generates scheduled runs of midnight EST every day

I tried to make the tip a little bit convoluted to show the inclusion of Sunday while skipping certain days.

### DIFF
--- a/docs/pipelines/process/scheduled-triggers.md
+++ b/docs/pipelines/process/scheduled-triggers.md
@@ -276,6 +276,9 @@ In the second schedule, **M-F 3:00 AM (UTC - 5) NC daily build**, the cron synta
 > [!IMPORTANT]
 > The UTC time zones in YAML scheduled triggers don't account for daylight savings time.
 
+> [!TIP]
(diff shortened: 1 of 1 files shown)


PRMerger17:
@Lampei : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x (approved):

steved0x:
@Lampei Thank you for that clarification, I may move that not just so we don't have two boxes in a row, and I will add a clarification up in the days of week section in the cron syntax as well. Thank you for your PR and sorry that your an into trouble.

#sign-off

ktoliver:
@steved0x Usually, I would make a few copy edits to this update before merging, but because you indicate that you'll be revising it, I'll merge it as-is. Thanks.

PR #9738, opened 2020-11-27
Allow *.dev.azure.com instead *dev.azure.com

markscholman:
This should be subdomain wildcard instead of *dev wildcard. This current implementation would mean hackeddev.azure.com eventually could be made available as urls


PRMerger12:
@markscholman : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):

PR #9811, opened 2020-12-10
Document 'force32bit' argument for NodeToolv0

DeeDeeG:
Hi folks, wanted to document this cool new feature that's been added to the NodeTool and UseNode tasks: https://github.com/microsoft/azure-pipelines-tasks/pull/13399

(Note: The feature is definitely live, as the Atom text editor has started using it in their CI: https://github.com/atom/atom/pull/21490. I wanted to make sure others knew about this neat feature that makes building 32-bit Node apps easy even on 64-bit Windows agents.)

Note: I formatted the documentation such that `force32bit` is above `checkLatest`, because `node-js.md` has a `[!TIP]` box below the table that is referring specifically to `checkLatest`. Otherwise I would have added it to the bottom of the table.

If there are any wording/formatting changes needed, or anything at all, feel free to let me know. Thanks.

### DIFF
--- a/docs/pipelines/tasks/includes/yaml/NodeToolV0.md
+++ b/docs/pipelines/tasks/includes/yaml/NodeToolV0.md
@@ -13,5 +13,6 @@ ms.technology: devops-cicd-tasks
 - task: NodeTool@0
   inputs:
     #versionSpec: '6.x' 
+    #force32bit: false # Optional
     #checkLatest: false # Optional
 ```
--- a/docs/pipelines/tasks/tool/node-js.md
+++ b/docs/pipelines/tasks/tool/node-js.md
@@ -35,6 +35,7 @@ None
 | Argument | Description |
 |----------|-------------|
 |`versionSpec`<br/> Version Spec | (Required) Specify which [Node.js version](https://nodejs.org/en/download/releases/) you want to use, using semver's version range syntax. <br/>**Examples**: `7.x`, `6.x`, `6.10.0`, `>=6.10.0` <br/>Default value: `6.x`|
+|`force32bit`<br/> Use 32 bit version on x64 agents | (Optional) Install the x86 version of Node.js on a 64-bit Windows agent. Only works on Windows.|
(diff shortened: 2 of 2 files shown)


PRMerger20:
@DeeDeeG : Thanks for your contribution! The author(s) have been notified to review your proposed change.

DeeDeeG:
Note: The reason that this only works on Windows is that it specifically only allows forcing `x86`, not generally forcing any arch string.

As you can see at the [Node.js.org downloads directory](https://nodejs.org/dist/latest-v15.x/), `x86` binaries are only built/available for Windows.

_(There is still a PR proposing to allow forcing any of the arches, but it has not been merged: https://github.com/microsoft/azure-pipelines-tasks/pull/8688)_

ktoliver:
@vijayma @steved0x Could you review the proposed update to this Pipelines article? Thanks.

DeeDeeG:
An alternate string for explaining this feature could be this, taken from the feature PR:

> Installs the x86 version of Node regardless of the CPU architecture of the agent.

_* (Not strictly accurate, as it implies this might work on, say, ARM Linux builds. It would error out there.)_

As opposed to what I have now:

> Install the x86 version of Node.js on a 64-bit Windows agent. Only works on Windows.

Or we could go with that but shorter:

> Install x86 Node.js on 64-bit Windows agents. Only works on Windows.

Or simply:

> Install x86 Node.js on 64-bit Windows agents.

ktoliver:
@steved0x and @vijayma - Could you review the proposed changes to this Pipelines article? Thanks.

steved0x:
@DeeDeeG Thank you for adding this!

steved0x:
#sign-off

steved0x (approved):

PRMerger8:
@steved0x: I'm sorry - only the author of this article, @vijayma, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #10008, opened 2021-01-15
Added rolloutStatusTimeout as available option

anderson-fachini:
I noticed that rolloutStatusTimeout was available using the classic pipeline editor (visual), but the option was not documented and I had some difficulty checking if, in fact, it was valid. Just [checking the code](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/KubernetesManifestV0/task.json) I was able to confirm that is was possible to use it.


PRMerger16:
@anderson-fachini : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
#sign-off

This looks right in the [task code](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/KubernetesManifestV0/task.json).

PRMerger7:
@juliakm: I'm sorry - only the author of this article, @azooinmyluggage, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #10025, opened 2021-01-19
Updating sprint / iteration limit

MS-alsulli:
It looks like we have an undocumented limit for sprints / iterations at 300 each per project. Adding this to the docs.


PRMerger15:
@MS-alsulli : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #10035, opened 2021-01-20
Fixes incorrect link title in "Analyze and report on code churn and code coverage"

Potherca:
In the "Related notes" section, a link titled "Code Churn" appears two times.

The first link seems correct, as it links to [Code Churn Excel Report
](https://docs.microsoft.com/en-us/previous-versions/azure/devops/report/excel/code-churn-excel-report).
The second link, however, links to [Code Coverage Excel Report](https://docs.microsoft.com/en-us/previous-versions/azure/devops/report/excel/code-coverage-excel-report).

Logic would dictate that the link title, therefore, should not be "Code Churn" but "Code Coverage".

This MR fixes that.


ktoliver:
Looks good; verified. Thanks for the fix, @Potherca.

PR #10069, opened 2021-01-26
Clarify parameter use in k8s manifest task

felixthiele:
This commit clarifies the use of the "manifest" parameter in the kubernetes manifest task of Azure Pipelines.

I recently tried out the kubernetes manifest task in my pipeline. Taking the old example (and explanation) of the manifest parameter, the pipeline would not find the manifest files to deploy. By using one path per line, it worked. So I assume that this is the way to do it (would correspond the "containers" parameter).

### DIFF
--- a/docs/pipelines/tasks/deploy/kubernetes-manifest.md
+++ b/docs/pipelines/tasks/deploy/kubernetes-manifest.md
@@ -6,7 +6,7 @@ ms.assetid: 31e3875c-c9ef-4c11-8b86-4b4defe84329
 ms.manager: atulmal
 ms.author: atulmal
 author: azooinmyluggage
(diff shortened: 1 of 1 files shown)


PRMerger14:
@felixthiele : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x (approved):

steved0x:
#sign-off

PRMerger20:
@steved0x: I'm sorry - only the author of this article, @azooinmyluggage, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #10072, opened 2021-01-27
Fixed a mistake on ooping through parameters

kmadof:
There is a mistake which causes displaying empty echo


PRMerger18:
@kmadof : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
#sign-off

Thanks @kmadof!

PR #10125, opened 2021-02-03
Fix test for working directory

Jaykul:
Use Test-Path to check the environment variable (as elsewhere) to be compatible with strict mode
Then pass the actual value of the environment variable to Test-Path to decide if we need to New-Item the directory


PRMerger18:
@Jaykul : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x (approved):

steved0x:
#sign-off

PR #10234, opened 2021-02-22
Update templates.md

NihinFaj:
When working with multiple jobs in one stage and one template file needs to be referenced, if the job name in the referenced template file is not empty, you get the error "job names must be unique within a stage". So it is good users know this too.


PRMerger19:
@NihinFaj : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
#sign-off

PR #10237, opened 2021-02-22
Fix for URL for .NET Core Windows Server Hosting bundle

tomtom1992:


PRMerger17:
@tomtom1992 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

docs/pipelines/apps/includes/prepare-aspnetcore-windows-vm.md
```diff
@@ -20,7 +20,7 @@ Invoke-WebRequest https://go.microsoft.com/fwlink/?linkid=848827 -outfile $env:t
 Start-Process $env:temp\dotnet-dev-win-x64.1.0.6.exe -ArgumentList '/quiet' -Wait
 
 # Install the .NET Core Windows Server Hosting bundle
-Invoke-WebRequest https://go.microsoft.com/fwlink/?LinkId=817246 -outfile $env:temp\DotNetCore.WindowsHosting.exe
+Invoke-WebRequest https://download.visualstudio.microsoft.com/download/pr/dff39ddb-b399-43c5-9af0-04875134ce04/1c449bb9ad4cf75ec616482854751069/dotnet-hosting-5.0.3-win.exe -outfile $env:temp\DotNetCore.WindowsHosting.exe
```
juliakm, line 23: Can you check if https://dotnet.microsoft.com/permalink/dotnetcore-current-windows-runtime-bundle-installer goes to the same place? If so, we should use the link that gets frequently updated instead of linking to a specific version.

juliakm (changes requested):

PR #10263, opened 2021-02-26
Rename branch folder name "releases" to "release"

GregTrevellick:
Appreciate this PR is somewhat subjective - submitting it on the basis that it's more standard to name release branch folder "release" rather than "releases". If I'm wrong then I've learnt something, no worries, if I'm right then we all win.

### DIFF
--- a/docs/pipelines/repos/includes/ci-triggers1.md
+++ b/docs/pipelines/repos/includes/ci-triggers1.md
@@ -16,10 +16,10 @@ You can control which branches get CI triggers with a simple syntax:
 ```yaml
 trigger:
 - master
-- releases/*
+- release/*
 ```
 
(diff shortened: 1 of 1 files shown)


PRMerger14:
@GregTrevellick : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x:
@GregTrevellick Thank you for your contribution, it is a good PR, but we have a lot of other places where we use `releases` as a folder for our branch filter examples, and if we merge this PR we will be inconsistent with our other examples. I haven't done a full sweep and I wouldn't be surprised if we mix and match with `release` and `releases`, but the first dozens of examples I found when doing a quick sweep used `releases`. I am filing a work item on myself to review, and if we change, I will change all of them so that we are consistent. Thank you again for your contribution.

Issue #10115, opened 2021-02-03
-bash: printf: -s: invalid option


xukai871105:
[Enter feedback here]
B64_PAT=$(printf -n":$MY_PAT" | base64)
-bash: printf: -s: invalid option

I think the right bash script is
printf "%s"":$MY_PAT"
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 53f99d2d-363b-d01d-039d-4fdc636fa2df
* Version Independent ID: f47330d5-e2de-36a4-e5b3-5bc77505b161
* Content: [Authenticate with personal access tokens - Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=preview-page)
* Content Source: [docs/organizations/accounts/use-personal-access-tokens-to-authenticate.md](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/organizations/accounts/use-personal-access-tokens-to-authenticate.md)
* Product: **devops**
* Technology: **devops-security**
* GitHub Login: @chcomley
* Microsoft Alias: **chcomley**

chcomley:
Hi @xukai871105, Thanks for your feedback. I've created a user story to address this. Once the updated article is published, this issue will automatically close. You can always @mention me if you have any further feedback or questions. Thanks again!

PR #10289, opened 2021-03-03
includes/PAT: Fix `Use a PAT` example in Bash

rico-chet:
In commit 3055c788bfaef1391e3d8690cc1630180989710b, a bad rebase seems
to have led to a broken shell example (a stray `-n` was added, probably
from `echo -n`). Also, since PAT can theoretically include characters
which have special meaning to `printf`, the example wasn't 100% correct
to begin with.

Fix both issues by removing `-n` and using a constant format string with
`printf`.

### DIFF
--- a/docs/repos/git/includes/personal-access-tokens.md
+++ b/docs/repos/git/includes/personal-access-tokens.md
@@ -135,7 +135,7 @@ On Linux or macOS, in Bash, you can enter:
  
 ```bash
 MY_PAT=yourPAT		# replace "yourPAT" with your actual PAT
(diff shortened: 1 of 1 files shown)


PRMerger9:
@rico-chet : Thanks for your contribution! The author(s) have been notified to review your proposed change.

vtbassmatt:
I'm no longer working on Azure DevOps.

steved0x (approved):

steved0x:
Fixed in another PR in response to https://github.com/MicrosoftDocs/azure-devops-docs/issues/10115

Issue #10426, opened 2021-03-26
Azure Artifacts documentation is missing out on the hard limit of 384000 bytes of Maven pom.xml files


j-tim:
I recently ran into the issue of the hard limit of 384000 bytes Azure Artifacts regarding the max size of a Maven `pom.xml` files.
This issue prevents me from resolving legitimate pom files like:

* [Lucene grandparent](https://repo1.maven.org/maven2/org/apache/lucene/lucene-solr-grandparent/8.8.1/)
* [Quarkus](https://repo1.maven.org/maven2/io/quarkus/quarkus-universe-bom/1.12.2.Final/)
* [Wildfly]( https://repo1.maven.org/maven2/org/wildfly/wildfly-parent/23.0.0.Final/)

from Maven central via my Azure Artifacts feed public upstream source. And probably many more big `pom.xml` files from open-source projects). 
The Azure Artifacts documentation is missing out on this hard limit of 384000 bytes on Maven `pom.xml` files 

See also: 
* https://developercommunity.visualstudio.com/idea/809379/remove-or-increase-the-hard-limit-of-384000-bytes.html

In my opinion, this should be documented [here](https://docs.microsoft.com/en-us/azure/devops/artifacts/reference/limits?view=azure-devops).

TimShererWithAquent:
@j-tim Thanks for your feedback; we appreciate it. To make sure we have all the information needed to address your issue, please let us know which article you’re referencing.

j-tim:
@TimShererWithAquent I don't understand your question. 
This issue about the limit is not documented. See my pull request to add it to the documentation.

The details about the issue with the hard limit (384000 bytes) on Maven project `pom.xml files itself can be found here: https://developercommunity.visualstudio.com/t/remove-or-increase-the-hard-limit-of-384000-bytes/809379.

ramiMSFT:
Ref: https://github.com/MicrosoftDocs/azure-devops-docs/pull/10427

PR #10427, opened 2021-03-26
#10426 Add hard size limit (384000 bytes) of Maven pom.xml files of Azure Artifacts to the documentation

j-tim:
Add hard size limit (384000 bytes) of Maven pom.xml files of Azure Artifacts to the documentation.

### DIFF
--- a/docs/artifacts/reference/limits.md
+++ b/docs/artifacts/reference/limits.md
@@ -22,6 +22,7 @@ Azure Artifacts is a highly-scalable artifact service. In the course of everyday
 * NuGet packages are limited to 500 MB.
 * npm packages are limited to 500 MB.
 * Maven packages are limited to 500 MB per file.
+  * Maven `pom.xml` files are limited to 384000 bytes.
 * Python packages are limited to 500 MB per file.
 ::: moniker range="<= tfs-2018 || azure-devops"
 * Universal Packages have been tested up to 1 TB and are recommended for managing large binary content.


PRMerger13:
@j-tim : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ramiMSFT:
@j-tim Thank you for your contribution, however your suggested change does not reflect the full picture of how Artifacts interact with POM files. We now take the uploaded POM, transform it, then apply the size limit to the transformed object. The transformed object is typically much smaller than the input.

j-tim:
@ramiMSFT apparently there is a hard limit on pom.xml files in Azure Artifacts. 
I and many other people ran into this issue: 

```
400 Bad Request - pom.xml file is too large; the limit is 384000 bytes
```

See also: https://developercommunity.visualstudio.com/t/remove-or-increase-the-hard-limit-of-384000-bytes/809379

electronerdian:
@j-tim I didn't see before that this PR actually predates the change ramiMSFT mentioned, which we made about a month ago. Are you still experiencing this problem?

jmyersmsft:
@j-tim I didn't see before that this PR actually predates the change ramiMSFT mentioned, which we made last month. Are you still experiencing this problem?

j-tim:
Hi @jmyersmsft yes I'm still experiencing this problem.

jmyersmsft:
@j-tim is this with a private package or one from Maven Central? If the latter, can you tell me the exact group/artifact/version you're having trouble with?

jmyersmsft:
If it's a private package, can you put the pom.xml in a zip file and tell me the size of the original POM and of the zip file?

j-tim:
@jmyersmsft it's not a private package. It's the grand parent pom of Lucene (used by Elasticsearch) for an example of the `pom.xml` file see: https://repo1.maven.org/maven2/org/apache/lucene/lucene-solr-grandparent/8.8.1/

jmyersmsft:
hmm... that should only be ~20KB after transformation, way under the limit. I'll have to check some things on my end.

jmyersmsft:
Well, this is embarrassing -- we never actually turned that feature on after we deployed the code. I have begun that process. Should reach your org within a few hours to a few days from now.

j-tim:
Thanks for the update @jmyersmsft. I'll try to run my project in the upcoming days and see whether this will fix the issue.
Just curious does this mean the hard limit of 384000 bytes for Maven pom.xml files will be gone?

jmyersmsft:
We hit a snag while enabling it last night, so it may be a few days before it's actually on. I'll try to remember to keep this thread updated.

I believe the limit is technically still there, but the internal compressed representation is what counts, so it should allow much, much larger actual pom.xml files. For example, org.apache.lucene:lucene-solr-grandparent:8.8.1's POM shrinks from a little under 400KB to ~20KB

j-tim:
That's great @jmyersmsft.

PR #10555, opened 2021-04-19
Added an environments section in the 'windows pipeline agents'

ericvruder:
added an environments section in the 'windows pipeline agents', and described the virtual machine resource tags parameters that can be used when setting up the agent. 

Done based on this stackoverflow post:

https://stackoverflow.com/questions/66870474/specifying-the-tags-on-the-environment-during-environment-creation


ericvruder:
I'm not quite sure how kubernetes resources work in the environment system - that's one of the reasons I did not add anything about them. There might be an issue though, since I am not sure if the parameters I added are only referencing virtual machines. It might be misleading since I don't mention that they only work for virtual machine resources.

steved0x (approved):

steved0x:
@ericvruder Thank you for your contribution!

#sign-off

Issue #10569, opened 2021-04-22
Broken yml example in the Deployment jobs


VyacheslavPritykin:
There is an incorrect yml example that:
- skips the execution of the StageB
- inserts an invalid value into the `myOutputVar` variable

Current non-working version:
```yml
stages:
- stage: StageA
  jobs:
  - deployment: A1
    pool:
      vmImage: 'ubuntu-16.04'
    environment: env1
    strategy:                  
      runOnce:
        deploy:
          steps:
          - bash: echo "##vso[task.setvariable variable=RunStageB;isOutput=true]true"
            name: setvarStep
          - bash: echo $(System.JobName)

- stage: StageB
  dependsOn: 
    - StageA
 
  # when referring to another stage, stage name is included in variable path
  condition: eq(dependencies.StageA.outputs['StageA.A1.setvarStep.RunStageB'], 'true')
  
  # Variables reference syntax differs slightly from inter-stage condition syntax
  variables:
    myOutputVar: stageDependencies.StageA.A1.outputs['A1.setvarStep.RunStageB']
  jobs:
  - deployment: B1
    pool:
      vmImage: 'ubuntu-16.04'
    strategy:                  
      runOnce:
        deploy:
          steps:
          - bash: echo $(myOutputVar)
```

Updated working version:
```yml
stages:
- stage: StageA
  jobs:
  - deployment: A1
    pool:
      vmImage: 'ubuntu-16.04'
    environment: env1
    strategy:                  
      runOnce:
        deploy:
          steps:
          - bash: echo "##vso[task.setvariable variable=RunStageB;isOutput=true]true"
            name: setvarStep
          - bash: echo $(System.JobName)

- stage: StageB
  dependsOn: 
    - StageA
 
  # when referring to another stage, stage name is included in variable path
  condition: eq(dependencies.StageA.outputs['A1.setvarStep.RunStageB'], 'true') # <---- modified
  
  # Variables reference syntax differs slightly from inter-stage condition syntax
  variables:
    myOutputVar: $[stageDependencies.StageA.A1.outputs['setvarStep.RunStageB']] # <---- modified
  jobs:
  - deployment: B1
    pool:
      vmImage: 'ubuntu-16.04'
    strategy:                  
      runOnce:
        deploy:
          steps:
          - bash: echo $(myOutputVar)
```

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

[...]

PR #10571, opened 2021-04-22
Fix yml example in deployment-jobs.md

VyacheslavPritykin:
#10569

### DIFF
--- a/docs/pipelines/process/deployment-jobs.md
+++ b/docs/pipelines/process/deployment-jobs.md
@@ -500,17 +500,11 @@ When you output a variable from a deployment job, referencing it from the next j
 stages:
 - stage: StageA
   jobs:
-  - deployment: A1
-    pool:
-      vmImage: 'ubuntu-16.04'
-    environment: env1
-    strategy:                  
-      runOnce:
-        deploy:
-          steps:
-          - bash: echo "##vso[task.setvariable variable=RunStageB;isOutput=true]true"
-            name: setvarStep
-          - bash: echo $(System.JobName)
+  - job: A1
+    steps:
+      - pwsh: echo "##vso[task.setvariable variable=RunStageB;isOutput=true]true"
+        name: setvarStep
+      - bash: echo $(System.JobName)
 
 - stage: StageB
   dependsOn: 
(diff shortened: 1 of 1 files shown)


PRMerger10:
@VyacheslavPritykin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

TimShererWithAquent:
@juliakm for awareness

juliakm:
@VyacheslavPritykin  Thank you for looking into this! Are you able to get your YAML to successfully run? I'm finding that StageB still gets skipped.

VyacheslavPritykin:
@juliakm Well, I actually tested it in a scenario, when the variable is set inside the non-deployment job and that works:

```yml
- job: A1
  steps:
    - pwsh: echo "##vso[task.setvariable variable=RunStageB;isOutput=true]true"
      name: setvarStep
    - bash: echo $(System.JobName)
 ```

I can confirm that if a variable is set inside a `deployment` job, then it still does not work, so it looks like there is an issue not only in the documentation but also in the pipeline engine.

I'll update the PR  to have a non-deployment job A1, so at least the documentation will have a valid example.

Moreover, the snippet from the documentation has another flaw (probably you've already found it): `Job B1: Environment is required.` (I didn't find it initially, because I needed a stage condition for a non-deployment job)

**Comment by PRMerger16** [comment]
@VyacheslavPritykin : Thanks for your contribution! The author(s) have been notified to review your proposed change.

**Comment by VyacheslavPritykin** [comment]
> I'll update the PR to have a non-deployment job A1, so at least the documentation will have a valid example.

I've updated and then I realized that it makes no sense since the page describes the Deployment Jobs and the difference in the syntax of consuming variables. We need to fix the initial version where a variable is set in the deployment job.

I've found how to fix the variable assignment in the second job:
```yml
myOutputVar: $[stageDependencies.StageA.A1.outputs['A1.setvarStep.RunStageB']]
myOutputVar: $[stageDependencies.StageA.A1.outputs['A1.setvarStep.RunStageB']]
```

but I'm still struggling with the valid condition expression

juliakm:
Thanks @VyacheslavPritykin!

#sign-off

PR #10692, opened 2021-05-10
change confusing terminology

shadargee1982:
Changed references to "Enable upstream behavior" since its on by default.  The allow external versions actually bypasses the behavior.  So I referred to the configuration as Allow external versions, and used the verb 'configure' where appropriate.


PRMerger16:
@shadargee1982 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ramiMSFT (approved):

ramiMSFT:
#sign-off

PR #10720, opened 2021-05-13
I lost some of the runs. Is there any way to get them back?  - paragraph

latudor:
While troubleshooting a case, i have stumbled upon the misunderstanding that can come out of the paragraph mentioned in the title.
I would like to make clear differentiation between the fact that runs that have been deleted by retention policies they cannot be recovered and the fact that if the pipeline deletion has passed a week it is also not possible to have it recovered.
Also, an important aspect to note is that pipeline deletions that have passed a week, must be manual, otherwise the retention policy not being able to recover gets trumped by the fact that only one week has passed. For this reason i have added the "manually" , in the phrase addressing this.

Please correct me if i am wrong.


PRMerger7:
@latudor : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
@latudor Thanks for your contribution! I'm checking on this with the product team.

juliakm:
Thanks @latudor. 

#sign-off

juliakm (approved):

PR #10929, opened 2021-06-15
Fixed the definitionId of the run retention examples

PabloL007:
Previously, the example scripts were setting the `definitionId` attribute of the `NewRetentionLease` object to `1` instead of using the predefined variable, which resulted in the request failing with the following response:
```json
{
	"$id": "1",
	"innerException": null,
	"message": "Build pipeline 1 was not found.",
	"typeName": "Microsoft.TeamFoundation.Build.WebApi.DefinitionNotFoundException, Microsoft.TeamFoundation.Build2.WebApi",
	"typeKey": "DefinitionNotFoundException",
	"errorCode": 0,
	"eventId": 3000
}
```
Switching this value to `$(System.DefinitionId)` will instead dynamically fill in the attribute with the definition ID of the running pipeline.

### DIFF
--- a/docs/pipelines/build/run-retention.md
+++ b/docs/pipelines/build/run-retention.md
@@ -61,7 +61,7 @@ This is similar to above, only the condition needs to change:
     script: |
       $contentType = "application/json";
       $headers = @{ Authorization = 'Bearer $(System.AccessToken)' };
-      $rawRequest = @{ daysValid = 365 * 2; definitionId = 1; ownerId = 'User:$(Build.RequestedForId)'; protectPipeline = $false; runId = $(Build.BuildId) };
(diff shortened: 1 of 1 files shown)


PRMerger16:
@PabloL007 : Thanks for your contribution! The author(s) have been notified to review your proposed change. @moswald

moswald:
@PabloL007 Oops! I could have sworn I went back and replaced my test hard-coded values with the system variables, but it looks like I missed one. Thanks!

I'm not actually sure how I'm supposed to do this. I think I just add the tag:

#sign-off

PabloL007:
Happy to help! It looks like someone else has to merge it though

moswald:
Yep. I believe the `ready-to-merge` tag acts like a bat-signal. Someone will come along.

PabloL007:
Makes sense, I've never seen a MicrosoftDocs maintainer and batman in the same room 😉 thanks!

PR #10945, opened 2021-06-17
Update AzureKeyVault arguments

50Wliu:
Update argument names to reflect the ones given in the YAML snippet.
Also, mark `runAsPreJob` as optional.

### DIFF
--- a/docs/pipelines/tasks/deploy/azure-key-vault.md
+++ b/docs/pipelines/tasks/deploy/azure-key-vault.md
@@ -79,10 +79,10 @@ When you want to access secrets:
 
 | Parameter | Description |
 | --------- | ----------- |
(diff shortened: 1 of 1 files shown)


PRMerger6:
@50Wliu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
@50Wliu Based on the task json [here](https://github.com/microsoft/azure-pipelines-tasks/blob/master/Tasks/AzureKeyVaultV2/task.json), I think the opposite it true. It looks like the descriptions don't match the updated task. 

Closing the PR and will update the task descriptions.

juliakm:
Just to follow-up. The issue was that we had documentation for V1 and V2 of the task in the same article. I'm adding an update now. Thanks for bringing this to our attention!

PR #10947, opened 2021-06-18
Not region but geography for Artifacts

rikat-ms:
Azure Storage account for Azure Artifacts can be created in Azure Geography, So, "Storage.{your region}" is not enough to allow communications in order to user Azure Artifacts.


PRMerger20:
@rikat-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):

PR #10971, opened 2021-06-23
Remove placeholder note

flcdrg:
I'm assuming this was a placeholder that was never updated, so I think it can be removed.


PRMerger18:
@flcdrg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):

PR #11196, opened 2021-08-13
switch to latestVersion which is the same text from the includes and …

MarkIannucci:
…therefore more likely to be accurate.

I noticed there's a conflict in the input description table from the yaml that is included above.  I assume the included one is more likely to  be accurate.


PRMerger9:
@MarkIannucci : Thanks for your contribution! The author(s) have been notified to review your proposed change.

steved0x (approved):

steved0x:
@MarkIannucci Thank you for your contribution. Task arguments and properties have a "friendly name" that is displayed in the task assistant, and then a "code" name, that is used in the YAML. This leads to some confusion, and we are working on an update to the tasks reference that makes this more clear. Thank you.

steved0x:
#sign-off

PRMerger13:
@steved0x: I'm sorry - only the author of this article, @RoopeshNair, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

Issue #10941, opened 2021-06-17
https://*.vsblob.visualstudio.net or https://*.vsblob.visualstudio.com


furmangg:
We're seeing this error in the logs

Error log :
[2021-06-16 22:11:47Z ERR  VisualStudioServices] GET request to https://vsblobprodcus3.vsblob.visualstudio.com/A71278a3c-bb80-492d-98c5-xxxxxxxxxxxx/_apis/connectionData?connectOptions=1&lastChangeId=-1&lastChangeId64=-1 failed. HTTP Status: Redirect

The docs say whitelist https://*.vsblob.visualstudio.net but that clearly is .com not .net. Should the docs be fixed? Or is there some nuance I'm not understanding?

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 425b68d1-4ba0-5657-4e09-f3aac361ea09
* Version Independent ID: f8044e9d-98e2-6e00-b513-a91e772b2ef8
* Content: [Deploy an Azure Pipelines agent on Windows - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-windows?view=azure-devops#for-organizations-using-the-devazurecom-domain)
* Content Source: [docs/pipelines/agents/v2-windows.md](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/pipelines/agents/v2-windows.md)
* Product: **devops**
* Technology: **devops-cicd-agents**
* GitHub Login: @steved0x
* Microsoft Alias: **sdanie**

TimShererWithAquent:
@furmangg - Thanks for your question. It looks like this one is already under investigation: /MicrosoftDocs/azure-devops-docs/issues/10645

If that issue doesn't cover your concerns, please reopen this one.

furmangg:
@TimShererWithAquent good find. Very similar issue with the section of doc right above the one I’m focused on. It seems they came to the same conclusion as me but the documentation isn’t fixed. Can we reopen this and ensure both sections get fixed?

Issue #10645, opened 2021-05-05
Whitelist of URLs is not correct for organizations with *.visualstudio.com domain


gintuoklis:
Whitelist of URLs is not correct for organizations with *.visualstudio.com domain

Our organization whitelisted all URLs for  *.visualstudio.com domain. 
But when pipeline tries to dowload artifacts, it tries to dowload from *.vsblob.visualstudio.net domain, which is part of another list. 

Don't know if there are more mistakes. Could you please review both list, and update accordingly?

![image](https://user-images.githubusercontent.com/16342128/117105109-a541c400-ad7d-11eb-859e-920d0755e661.png)

[Enter feedback here]

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 425b68d1-4ba0-5657-4e09-f3aac361ea09
* Version Independent ID: f8044e9d-98e2-6e00-b513-a91e772b2ef8
* Content: [Deploy an Azure Pipelines agent on Windows - Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/v2-windows?view=azure-devops#im-running-a-firewall-and-my-code-is-in-azure-repos-what-urls-does-the-agent-need-to-communicate-with)
* Content Source: [docs/pipelines/agents/v2-windows.md](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/pipelines/agents/v2-windows.md)
* Product: **devops**
* Technology: **devops-cicd-agents**
* GitHub Login: @steved0x
* Microsoft Alias: **sdanie**

TimShererWithAquent:
@gintuoklis - Thanks for your question. It looks like you're working through an issue with your scenario or implementation, rather than an issue with the documentation. Please consider these resources:

- [Azure DevOps Services](https://azure.microsoft.com/support/devops/) 
- [Azure DevOps Support Bot](https://azuredevopsvirtualagent.azurewebsites.net/) 
- [Azure DevOps on Stack Overflow](https://stackoverflow.com/questions/tagged/azure-devops)

gintuoklis:
sorry, but the issue is with documentation. Your lists are not correct. If you follow your documenation for *.visualstudio.com domain, and add all addresses to firewall exceptions, your agent still cannot download blobs. THerefore, i opened the request so you could review/update those lists.

steved0x:
[...]

PR #11209, opened 2021-08-16
vsblob.visualstudio.net -> vsblob.visualstudio.com

rikat-ms:
One of my customers encountered the issue due to the wrong URL. Based on the following Issues, the document should be corrected as soon as possible to prevent the other users from encountering the same issue.
https://github.com/MicrosoftDocs/azure-devops-docs/issues/10645
https://github.com/MicrosoftDocs/azure-devops-docs/issues/10941


PRMerger14:
@rikat-ms : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #11237, opened 2021-08-20
Add credentials to find/install commands

samfoley88:
I have a system where I don't have to provide credentials but that system is my development box so I think on a system just pulling from the repo you will always need to provide them.
If somebody can correct me great but this new update agrees with the medium article referenced for this guide.

### DIFF
--- a/docs/artifacts/tutorials/private-powershell-library.md
+++ b/docs/artifacts/tutorials/private-powershell-library.md
@@ -271,17 +271,19 @@ We now have a private repository within Azure Artifacts that we can push our Pow
     ```powershell
         Get-PSRepository
     ```
+    > [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger9:
@samfoley88 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ramiMSFT:
@samfoley88 Thank you for your feedback! The credentials are already provided in the previous steps when registering "PowershellAzureDevopsServices"

samfoley88:
Thanks for the response.

They are but they don't get passed through when provided there for the later commands. Running that sequence on a clean Windows install returns an error wihout the credentials and still requires credentials on the later steps as I've updated in the docs.

Not at my computer but will update with the error and version info for replication when I am.

I'd be very happy to be proven wrong on this as it's a big pain!

PR #11251, opened 2021-08-25
Clarified limitation between VMSS and DevOps

JPGRodrigues:
A VMSS should be associated to a single Azure DevOps pool.
Associating the same VMSS to multiple Azure DevOPs pool is not resulting in any error, but the VMSS will have strange behavior because it may receive concurrent/antagonic requests from Azure DevOps


JPGRodrigues:
When trying to associate the same VMSS to multiple Azure DevOps pool we are getting the correct error: This virtual machine scale set is already in use by pool <poolname>
But anyway, the documentation should clearly state that this is not a supported scenario and help quickly help consumers not to try this configuration. 
There are support tickets being opened to clarify this situation.

PR #11258, opened 2021-08-27
Add missing "pat" to "patterns" in download schema

winstliu:
😃


juliakm:
#sign-off

@50Wliu Yikes, thanks for catching that!

Issue #11289, opened 2021-09-02
Section "Add Mermaid diagrams to a Wiki page" is displayed for unsupported products


Arturace:
In the version drop down list, when a user selects a "Azure DevOps Server", the "Add Mermaid diagrams to a Wiki page" section should **not** be displayed. It is misleading and makes selecting your version rather useless.

![image](https://user-images.githubusercontent.com/36833825/131898177-c437e42a-2e7f-4669-a8f6-c8249b5367f8.png)

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 29f1e43f-a069-dd88-7fc7-9cc497fb94f0
* Version Independent ID: c261afa7-6eb7-b3f0-9f92-6d2f3c5ac336
* Content: [Markdown syntax for wikis in Azure DevOps - Azure DevOps](https://docs.microsoft.com/en-us/azure/devops/project/wiki/wiki-markdown-guidance?view=azure-devops-2020)
* Content Source: [docs/project/wiki/wiki-markdown-guidance.md](https://github.com/MicrosoftDocs/azure-devops-docs/blob/master/docs/project/wiki/wiki-markdown-guidance.md)
* Product: **devops**
* Technology: **devops-collab**
* GitHub Login: @chcomley
* Microsoft Alias: **chcomley**

chcomley:
We've just published this update, which should go live shortly! Thanks so much for your feedback.

PR #11298, opened 2021-09-07
Fix mermaid.js availability in wiki

Arturace:
Fixing #11289.

I changed the moniker range on the corresponding section. I was not able to make monikers work locally, but I checked similar uses of this monicker in some other files, so it should work. If someone has a resource I can read up on making monikers work locally, I'd appreciate it.

I removed the note indicating that mermaid.js in the markdown is only available for the devops services, as it is redundant if the section is never shown if the a server version is selected.


PRMerger8:
@Arturace : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):

PR #11461, opened 2021-10-08
Update jenkins-queue-job.md

kripa-2019:
The steps in the document will not work because of the dependency on the Team Foundation Server Plug-in which is now deprecated per : https://issues.jenkins.io/browse/INFRA-2751


ktoliver:
@RoopeshNair Note that the PR appears to report an issue instead of proposing an update, so it's not ready to merge. Thanks.

PR #11493, opened 2021-10-19
TFVC will not receive future investment

SamGrantham:
Per the conversation around a recent Design Change Request targeting TFVC, there will be no future investment in this system (baring security related issues) because of this we need to *AT LEAST* update the page we offer for helping customers make a decision on which version control system to use.  Rather than try to determine the exact status of TFVC, I emphasized that Git is the system that is receiving investment hopefully making clear that this should be your choice, and that you should start moving away from TFVC to Git.  Hopefully this is strong enough language to help people choose and to sight from support when requests are made for new work or bug-fix.

### DIFF
--- a/docs/repos/tfvc/comparison-git-tfvc.md
+++ b/docs/repos/tfvc/comparison-git-tfvc.md
@@ -21,7 +21,7 @@ and [Team Foundation Version Control](./what-is-tfvc.md) (TFVC).
 <a name="tfvc_or_git_summary"></a>
 ## Which version control system should I use?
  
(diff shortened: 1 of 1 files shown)


PRMerger16:
@SamGrantham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

MaximMalygin:
For me It is not good idea notify people to use TFVC, if customer has huge repo, they should use TFVC, git is not best choice if repo is the really huge
cc: @vijayma

PR #11565, opened 2021-11-03
Fix receiving signals in start.sh

zoltanhomoki:
The current start.sh script is prepared to receive and handle signals like TERM and do some cleanup work (such as removing the agent from the portal).
However it runs the run.sh as a foreground process so even if it receives the signals it has to wait for the run.sh (which is the agent itself) to finish. But it won't finish as the signal is not propagated to the child.
An easy fix is to run the run.sh as a background process and wait for it at the end. In this case the start.sh can receive the signals properly and do the cleanup job. It's been tested and works fine.

Also I've updated the agent version to latest in the Dockerfile.

### DIFF
--- a/docs/pipelines/agents/docker.md
+++ b/docs/pipelines/agents/docker.md
@@ -226,7 +226,7 @@ Next, create the Dockerfile.
       && rm -rf /var/lib/apt/lists/*
 
     ARG TARGETARCH=amd64
-    ARG AGENT_VERSION=2.185.1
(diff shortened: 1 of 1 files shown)


PRMerger8:
@zoltanhomoki : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

steved0x (approved):

steved0x:
#sign-off

PR #11646, opened 2021-11-23
Update instructions to run self-hosted agent in Docker container for Linux

AndreyIvanov42:
Running an agent in a container has its own rules.

One of these rules is [this](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#create-ephemeral-containers).

The changes take into account aspects of the agent's behavior inside container.
The logic for determining and downloading the latest version of the agent has been moved to the stage of starting the container. This allows us to run the latest version of agent and corresponds to the Windows section.

Testing using ubuntu:18.04 docker image and agent v2.194.0, v2.195.0


PRMerger4:
@AndreyIvanov42 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #11838, opened 2022-01-13
Basic Authentication does not use PATs

PagatHackimo:
As stated in the docs above IIS Basic Authentication breaks PAT Authentication. Therefore a user:PAT combination does not work and should be user:password as you are user Basic Authentication and not the "PAT Basic Auth".


PRMerger12:
@PagatHackimo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):

PR #11968, opened 2022-02-13
Job Reuse typo?

john-landgrave:


PRMerger-2:
@john-landgrave : Thanks for your contribution! The author(s) have been notified to review your proposed change.

john-landgrave:
Sorry, I used the github.dev editor so I'm not sure if there was a PR template that wasn't followed? I was reading the documentation on Job Reuse in Azure Pipelines and it appeared that there was an inconsistency in the section I've amended. If my understanding of the paragraph preceding the code block that I've edited is incorrect, then I would propose that there is another change that needs to be made to clarify that?

Essentially, comparing the two code blocks in this section, it looks like the `Ubuntu` job name was removed, but the `Windows` one wasn't, which seems like an issue considering the language: "When working with multiple jobs, remember to remove the name of the job in the template file, so as to avoid conflict."

PR #11991, opened 2022-02-21
Update expressions.md

dciborow:
Provide a useful example of how to use convertToJson with Bash. See this issue here. https://stackoverflow.com/questions/57264556/prevent-quotes-from-disappearing-in-json-in-between-azure-devops-tasks/70967584?noredirect=1#comment125795408_70967584

### DIFF
--- a/docs/pipelines/process/expressions.md
+++ b/docs/pipelines/process/expressions.md
@@ -217,6 +217,9 @@ parameters:
  
 steps:
 - script: |
+    # Based on Solution Provided by Jochen van Wylick
+    # This is required to properly format the JSON
+    # https://stackoverflow.com/a/57388706/1341806
     json=$(echo "${{ convertToJson(parameters.listOfValues) }}" | \
       sed -e 's/^  /  "/g' | \
       sed -e 's/: /": "/g' | \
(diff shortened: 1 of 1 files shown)


PRMerger9:
@dciborow : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger4:
@dciborow : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

PRMerger16:
@dciborow : Thanks for your contribution! The author(s) have been notified to review your proposed change.

dciborow:
Even better fix provided. The correct way here is to use single quotes instead of double qoutes.

juliakm:
Thanks @dciborow! 

#sign-off

SimonAlling:
> The correct way here is to use single quotes instead of double qoutes.

No, it just happens to work more often because double quotes are more common than single quotes in JSON. The real solution is to pass the JSON to the program as _data_ instead of using it as part of the _source code_ of the program. Please see [my comment](https://github.com/MicrosoftDocs/azure-devops-docs/issues/11983#issuecomment-1125092780) on #11983 as well as my suggested improvement in #12312.

PR #12002, opened 2022-02-24
configurationString value should not contain colon

janssen-io:
The configuration string expects one or more `-<key> <value>` pairs. Adding a colon behind the key makes it part of the key. This prevents the configuration from actually being set.


ktoliver:
#label:"aq-pr-triaged"

steved0x (approved):

steved0x:
#sign-off

PRMerger8:
@steved0x: I'm sorry - only the author of this article, @azooinmyluggage, can sign off on your changes. But we do have an exception process - if you are on the Microsoft content or product team for this product area, you can ask the PR review team to review and merge it by sending mail to the techdocprs alias.

PR #12053, opened 2022-03-08
Add back section on default .git exclusion

50Wliu:
I've confirmed this through multiple tests.
Azure Artifacts will exclude the `.git` folder by default if you don't have an .artifactignore folder, so this section (removed in 395ab7aa45c1df2ddf9acffd77502a6e0199f84d) should be added back with a few modifications.

It should probably also appear in the ["Publish artifacts"](https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/pipeline-artifacts?view=azure-devops&tabs=yaml#publish-artifacts) article, given that this behavior happens **without** an .artifactignore file, so people might not even think to look there.


PRMerger17:
@50Wliu : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ramiMSFT (approved):

ramiMSFT:
#sign-off

PR #12075, opened 2022-03-16
Update run-retention.md

JFolberth:
The `definitionId` in the example is hardcoded to 1. This will not work for users copying and pasting and should be dynamic. The other examples have this value set to `$(System.DefinitionId)`


PRMerger20:
@JFolberth : Thanks for your contribution! The author(s) have been notified to review your proposed change. @moswald

Court72:
Hi @moswald 

Can you join the Microsoft Docs org so that you can approve these proposed changes?

Thanks!

ShannonLeavitt:
@steved0x - The ms.author for this article doesn't seem to be part of the Microsoft Docs org. If we don't get a response from him, would you be in a position to review and approve the changes here?

moswald (approved):

PR #12137, opened 2022-03-30
Missing parent for job in user defined variables

jontreynes:
Missing parent for job in user defined variables documentation when describing how to pass a variable to another stage


Court72:
#label:"aq-pr-triaged"

PR #12394, opened 2022-06-13
Update decide-between-using-local-server-workspace.md

kerimeredith:
typo - changed "It may be also be preferable " to "It may also be preferable "
(there is not a 2nd change)


PRMerger13:
@kerimeredith : Thanks for your contribution! The author(s) have been notified to review your proposed change. @MaximMalygin

PR #12437, opened 2022-06-26
Variables are not available by default to the entire stage

flcdrg:
As mentioned further down in the document, variables are not by default available to other jobs in the current stage, so I think using the word 'stage' here is incorrect. Variables are available by default to other tasks in the current job, so 'job' seems like the correct term.


PRMerger8:
@flcdrg : Thanks for your contribution! The author(s) have been notified to review your proposed change.

juliakm:
Thanks @flcdrg!

#sign-off

juliakm (approved):

PR #12557, opened 2022-08-05
Fix URL and sample code in Note section

pnwguy:
URL was pointing to get-work-item instead of get-user page, and the sample html code was incomplete - it was missing the end tags.


PRMerger19:
@pnwguy : Thanks for your contribution! The author(s) have been notified to review your proposed change.

chcomley (approved):
Thanks so much @pnwguy!

PR #12644, opened 2022-08-31
Update prepare-permissions.md

akoppanur:
Please add an image to choose the Organization for which the PAT is being created for Azure DevOps Server versions.
Currently there is an active bug due to which only PAT for "All accessible Organizations" works. Here is the Bug: https://dev.azure.com/mseng/AzureDevOps/_workitems/edit/1942506 

Customer mentioned that the documentation does not indicate to choose PAT for "All accessible Organizations".

Actually, PAT should work for the specific collection but due to the active bug, for now, the workaround is that customers have to create PAT for "All accessible Organizations".


Court72:
@steved0x 

Can you review the proposed changes? If approved, the new image will need to be moved into an internal repo and renamed to follow the file naming guidelines.

#label:"aq-pr-triaged"

steved0x:
I am closing this and moving it to a PR in the private repo.

PR #12737, opened 2022-09-29
Update access-levels.md

ManojNair:
Early Adopter licence type need to be mentioned in the public docs if internal only


ManojNair:
Ignore the change request

Issue #12799, opened 2022-10-20
Remove duplicate entry


theBNT:
The page lists `https://app.vssps.visualstudio.com` twice

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: dc5a4cfa-aa84-9f0f-55f4-71ec4861c070
* Version Independent ID: d621620a-c924-03b1-8b43-f99c33e58bb8
* Content: [Allowed address lists and network connections - Azure DevOps](https://learn.microsoft.com/en-us/azure/devops/organizations/security/allow-list-ip-url?view=azure-devops&tabs=IP-V4)
* Content Source: [docs/organizations/security/allow-list-ip-url.md](https://github.com/MicrosoftDocs/azure-devops-docs/blob/main/docs/organizations/security/allow-list-ip-url.md)
* Service: **azure-devops**
* Sub-service: **azure-devops-security**
* GitHub Login: @chcomley
* Microsoft Alias: **chcomley**

PR #12800, opened 2022-10-20
this removes the duplicate URL to fix issue #12799

theBNT:
The host `https://app.vssps.visualstudio.com` was explicitly stated twice, which makes it redundant and it can be removed.

I deliberately left it in once though (because one could argue its matched by `https://*.vssps.visualstudio.com`), so it shows up in searches. Also it depends on the firewall implementation if wildcard works as expected.


Court72:
@chcomley 

#label:"aq-pr-triaged"

chcomley (approved):
#sign-off

PR #12808, opened 2022-10-20
Document that task commands unescape their value

schra:
Here
https://github.com/microsoft/azure-pipelines-task-lib/blob/e69409b1c9b038296e8d67954adac50840d9dec6/node/taskcommand.ts#L93-L97 you can see that the official ADO pipeline tasks escape their data that they pass after the `]` of a `##vso` command.

As far as I can tell, this behavior is not yet documented at all. Especially in the context of `task.setvariable` it should be mentioned how to set variables that contain newlines.


juliakm:
Thanks!

#sign-off

PR #12848, opened 2022-11-03
correcting "daylight saving time"

raydixon:
correcting `daylight savings time` to `daylight saving time`


Jak-MS:
@raydixon - thanks for submitting this correction!

PR #12908, opened 2022-11-21
fix condition for Docker cache save task

jennings:
The cache should only be populated if the build was neither canceled nor failed, but the previous example would overwrite the cache when the build failed.

This PR changes the example's condition to be:

```
and(
  not(canceled()),
  not(failed()),
  ne(variables.CACHE_RESTORED, 'true')
)


Court72:
@ramiMSFT 

#label:"aq-pr-triaged"

ramiMSFT (approved):

ramiMSFT:
Thanks for your contribution!

#sign-off

PR #12927, opened 2022-11-28
Uses v4 of the Azure DevOps Extension Tasks

jessehouwing:
The Azure DevOps Extension tasks have been upgraded to Node 16 and I deemed the requirement of a new Node version enough to up the major versio of the extension.


chcomley (approved):
#sign-off

PR #13017, opened 2023-01-08
Fix typo in upstream-sources.md

ericmutta:
Change this:

> Using scopes instead of upstream sources **limit** your private package consumption to those with the `@scope` prefix e.g. `@fabrikam/core` but **enable** you...

To this:

> Using scopes instead of upstream sources **limits** your private package consumption to those with the `@scope` prefix e.g. `@fabrikam/core` but **enables** you...


Court72:

ericmutta:
Many thanks @Court72 for following up on this and my other recent PRs! :+1:

PR #13018, opened 2023-01-09
Fix typos in feed-permissions.md

ericmutta:
Change this:

> By default, **every users** in the same organization have the permissions...

To this:

> By default, **users** in the same organization have the permissions...

Change this:

> Below are the **four** different access levels for a feed...

To this (because I count more than four levels, so it's probably safer to avoid mentioning the number of levels just in case they change again in the future):

> Below are the different access levels for a feed...


Court72:

PR #13113, opened 2023-02-10
Fix typo (securing -> security)

bendews:
Very minor typo fix - a little pedantic but caused me to have a double-read of the paragraph so I'm sure it has for others as well.


Court72:
@bendews - Thanks for that fix!

PR #13149, opened 2023-03-01
Fixed resourceType value and added new example

AdrianSanguineti:
The `resourceType` value specified for `virtualMachine` in  all the examples had the wrong casing, that causes deployment failures when the examples are used verbatim. This update should hopefuly save some folks the frustrations I went through.

The first tag filter example has also been replaced with a new `resourceName` parameter example as the existing example looked redundant when there is an entire section about resource tags which followed next.


Court72:
@juliakm

#label:"aq-pr-triaged"

PR #13154, opened 2023-03-02
IIS configuration docs only apply to ADO Server

akanieski:
IIS configuration docs only apply to ADO Server


chcomley (approved):
#sign-off

PR #13156, opened 2023-03-02
Update restrict-invitations.md

AlexandraRamirezMS:
Adding details of how to add users when the policy is enabled Confusing as the Add button will not show for the Project and Team Admins even when policy is enabled


chcomley (approved):

chcomley:
#sign-off

PR #13211, opened 2023-03-27
Remove dollar sign from command example in CLI doc

incarnate:
The "Command usage" example has a $ at the start - the other examples do not have this and using the copy button copies the $ to the clipboard.


chcomley (commented):
Thanks @incarnate

chcomley (approved):

chcomley:
#sign-off

PR #13220, opened 2023-03-31
Update service-principal-managed-identity.md

bjompen:
Fixed a link that was wrong.


bjompen:
Added a spelling error fix as well.
loorganization -> organization

ShannonLeavitt:
Looks good; I've tested the new link. Thanks @bjompen

PR #13254, opened 2023-04-17
Fix typo in service-principal-managed-identity.md

teesana:


Court72:
Looks good; thanks, @teesana

PR #13293, opened 2023-05-05
Update multi-repo-checkout.md (Triggering repository has to be explicitly checked out)

arkoppan:
On Line 272, removed this part of the line ", unless you explicitly override the behavior during checkout." Because the triggering repository is not automatically checkout (an internal bug was created and this information is tested and verified with PM).   

In the Yaml, added lines 297 to 300, so that the information in the Table that follows will still be applicable.

### DIFF
--- a/docs/pipelines/repos/multi-repo-checkout.md
+++ b/docs/pipelines/repos/multi-repo-checkout.md
@@ -294,6 +294,10 @@ resources:
     trigger:
     - main
     - release
+  steps:
+  - checkout: self
+  - checkout: A
+  - checkout: B
 ```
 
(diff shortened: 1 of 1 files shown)


arkoppan (commented):
I actually needed to make the first change in Line 299 and not 272. Line 272 is unchanged.
I created a new PR for this change to line 299: https://github.com/MicrosoftDocs/azure-devops-docs/pull/13294

arkoppan:
Please disregard this PR. Please see new PR: https://github.com/MicrosoftDocs/azure-devops-docs/pull/13295

PR #13294, opened 2023-05-05
Update multi-repo-checkout.md 

arkoppan:
-Updated line 299 by removing this part ", unless you explicitly override the behavior during checkout." Because the triggering repository is not automatically checkout (an internal bug was created and this information is tested and verified with PM).

- No changes in line 272 as mentioned in this PR: https://github.com/MicrosoftDocs/azure-devops-docs/pull/13293
- missed to make this change to line 299 in the previous PR: https://github.com/MicrosoftDocs/azure-devops-docs/pull/13293

PR #13295, opened 2023-05-05
Update multi-repo-checkout.md

arkoppan:
1. In the Yaml, added lines 297 to 300, so that the information in the Table that follows will still be applicable.

2. Modified original line 299 (now 303). Removed this part of the line ", unless you explicitly override the behavior during checkout". Because the triggering repository is not automatically checkout (an internal bug was created and this information is tested and verified with PM).

Please disregard/cancel, the previous 2 PRs:  

https://github.com/MicrosoftDocs/azure-devops-docs/pull/13293 https://github.com/MicrosoftDocs/azure-devops-docs/pull/13294

### DIFF
--- a/docs/pipelines/repos/multi-repo-checkout.md
+++ b/docs/pipelines/repos/multi-repo-checkout.md
@@ -294,9 +294,13 @@ resources:
     trigger:
     - main
     - release
+steps
+- checkout: self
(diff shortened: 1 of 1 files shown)


steved0x:
#sign-off

PR #13323, opened 2023-05-11
Fix bash snippet for cloning with a PAT

mathieu-lemay:
The example provided is totally wrong. Used as is, it doesn't work.

The main problem is that the comment tells the user to prefix their PAT with a  `:`, but then, on the next line, there is also a `:` added in the `printf`.  This ends up generating a wrong base64.

The printf is also completely wrong. No space between `"%s"` and `":$MY_PAT"`,  means it's just one string and no 2nd argument for the `"%s"`, making the  placeholder unused. The proper form is to have the `:` in the format string and  passing "$MY_PAT" as an argument.


Jak-MS:
@vijayma 
- Can you review this PR?
- IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

steved0x (approved):

steved0x:
#sign-off

PR #13463, opened 2023-07-07
Update conditions.md

akoppanur:
1. In line 129, Stage 1 does not have dependsOn: stage2. By default Stage 2 depends on stage1. So, in the below statement, it should be 'stage2' depends on 'stage1'

---------------
In this pipeline, stage1 depends on stage2. Job B has a condition set for it.
------------------

2. In Line 146, the 'Stage2' has Job B. So in the below statement, the job in 'stage2' should be job 'B' and not 'A'

---------------
If you queue a build on the `main` branch, and you cancel it while `stage1` is running, `stage2` *won't* run, even though it contains a job `B` whose condition evaluates to `true`.
 ---------------


juliakm:
@akoppanur Thanks for your detailed edits! It's contributors like you that make our documentation better. I appreciate your PR. 

#sign-off

PR #13520, opened 2023-07-27
Update about-projects.md

JFolberth:
Adding clarity as situations where customers are adamant one project is the only recommended approach based on current document wording.


chcomley (approved):

chcomley:
#sign-off

PR #13593, opened 2023-08-27
review-pull-requests.md: Fixed sample for three-dot-diff in PRs

Niklas-Peter:
It would not be useful if in a PR only the changes from master (F, G) would be shown. Instead, we want to see only the changes from the feature branch (C, D, E).

### DIFF
--- a/docs/repos/git/review-pull-requests.md
+++ b/docs/repos/git/review-pull-requests.md
@@ -166,7 +166,7 @@ For example:
 A---B---F---G   master
 ```
 
-Pull Requests: ```git diff branch...master``` will produce only F, G commits. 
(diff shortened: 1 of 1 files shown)


steved0x:
@bjanousek could you review the PR please?

bjanousek:
@Niklas-Peter - the change proposed is incorrect. If we want to show to user that the diff displayed will be C. D, E then we need to change the order of the branches too.

CC @steved0x

bjanousek (changes requested):
The change is incorrect, If we want to show the result from the command the changes done in the branch, then the command must be modified too.

steved0x (approved):

steved0x:
#sign-off

PR #13677, opened 2023-10-02
Move a comma to make a sentence make sense.

bjompen:
The comma was misplaced. Moving it makes the sentence make more sense.


Jak-MS:
@bjompen - thanks for your edit!

PR #13695, opened 2023-10-12
Update logging-commands.md

DryadYao000:
Hello Team, during the tests, the output is different from the showing in the wiki, especially the usage of the "nested lists". Per the sharing from EEE, this difference is as expected. My customer wants to know whether the difference can be described in the documentation.


DryadYao000:
Hello team,

PR #13703, opened 2023-10-13
Trigger child pipeline for the same branch as the parent pipeline

akoppanur:
Had a long running case where, when parent is triggered for different branches, the child pipeline would run only for one of the branches for which master is triggered. 

Added lines 91 to 106

### DIFF
--- a/docs/pipelines/process/pipeline-triggers.md
+++ b/docs/pipelines/process/pipeline-triggers.md
@@ -88,7 +88,8 @@ resources:
         exclude:
         - releases/old*
 ```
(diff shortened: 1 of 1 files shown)


steved0x (approved):

steved0x:
@akoppanur Thank you for this update. I made a small change, and it triggered off the licensing bot. Could you reply with a PR comment as described above? Then I will merge this.

#sign-off PR Merger team please merge this when the licensing is accepted.

Jak-MS:
@akoppanur - after you've signed the CLA (see above for instructions), you'll need to let @steved0x  or me know so we can sign off again and get this merged. thanks

PR #13709, opened 2023-10-16
Update permissions-access.md - typo

PLsergent:


Jak-MS:
@PLsergent - thanks for this fix!

PR #13787, opened 2023-11-15
Fixed broken URLs for Ubuntu images

andreas-nesheim:
URLs for `ubuntu-latest` and `ubuntu-20.04` were broken. Fixed these since it looks like some file structure was renamed.

PR #13796, opened 2023-11-16
Fix/Broken URLS for hosted agents

kek-Sec:
related to #13787 but also fixes windows links


ShannonLeavitt:
Looks good. Thanks @kek-Sec

PR #13944, opened 2024-01-19
Change location of imposed limits on Azure Templates

AbhishekMarda:
The imposed limits (specially information about max 100 YAML files) is currently located at the very end of the last page on templates can very easily be missed. Such limits can often influence design decisions for the end user (as it did for my team). This information should be available at the first page so that customers can understand their restrictions while going in to the product. 

A topic like template expressions may or may not be opened, and especially may not be scrolled to the end. 

This PR shifts the information to the main page on templates to solve this.


Court72:
@juliakm

#label:"aq-pr-triaged"

juliakm:
Thanks @AbhishekMarda!

#sign-off

PR #13971, opened 2024-01-31
Fix markdown link

rogerthatdev:
Replace `[...]` with `(..)`


Jak-MS:
@rogerthatdev - thanks for this fix!

PR #14042, opened 2024-03-06
Updating azure-key-vault.md for tutorial to work and formatting.

johnmart82:
## Updates made.

1. Adding a note to the top of the page to advise readers that they will need to apply for parallel job as there is not one allocated by default to new Azure DevOps organizations. Currently the tutorial will fail with an error stating that there are no parallel jobs and provide the link. This is not a great reader experience and I found frustrating when following the tutorial.
2. Moving the warning about this not following security best practices from the bottom of the tutorial to the top in order to make it more prominent. This will help make sure that the reader understands that they need to do additional reading and planning to use this in a production setting.
3. Currently the tutorial will not work as the service principal (SP) creation does not actually get used by Azure DevOps because there is nothing telling readers how to use it to setup a service connection once the SP is created. I have updated this flow now so that the reader will be guided through assigning rights to the secrets using an access policy for the service connection created when the key vault helper is used at the start of the pipeline creation process.
4. I have updated "Save" to "Save and run" which is the option now used on pipeline creation.
5. Removed the classic tab for new pipeline creation as the option is no longer available on the new pipeline creation workflow in Azure DevOps.
6. Removed the YAML header in the New Pipeline section as it is no longer needed now that the classic tab has been removed.

### Question
The documentation page currently states that Key Vaults which use RBAC are not supported for use with this Key Vault task (Line 191). I have done some testing and if the Key Vault is switched to RBAC and the service connection principal is given the "Key Vault Secrets User" role assignment it will function the same as if the access policy was in use. Can we check with the product team and see if this is still the case or whether the RBAC capability is now supported and the docs just need to catch up? It would be a good addition to have here given the legacy nature of the key vault access policies.

### DIFF
--- a/docs/pipelines/release/azure-key-vault.md
+++ b/docs/pipelines/release/azure-key-vault.md
@@ -212,8 +212,6 @@ Don't save or queue your pipeline just yet. We must first give our pipeline the 
 
 ## Set up Azure Key Vault access policies
 
-In order to access our Azure Key Vault, we must first set up a service principal to give access to Azure Pipelines. Follow [this guide](/azure/active-directory/develop/howto-create-service-principal-portal#register-an-application-with-azure-ad-and-create-a-service-principal) to create your service principal and then proceed with the next steps in this section.
-
 1. Navigate to [Azure portal](https://azure.microsoft.com/).
 
 1. Use the search bar to search for the key vault you created earlier.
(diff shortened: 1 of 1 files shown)


Jak-MS:
@ramiMSFT 
- Can you review this PR?
- IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"

johnmart82:
@ramiMSFT Any chance that you can re-open this and take a look? Many thanks.

ramiMSFT:
Thanks for your contribution. However, this article has recently been reviewed and overhauled to update the setup and authentication process. Please check out the latest version. 
Classic is still part of the pipeline experience; you just have to switch to the native experience.

PR #14075, opened 2024-05-02
fix encoding the token

parafoxx:
Only the credentials have to be base64 encoded. The rest of the header text needs to stay in plaintext as can be seen in the powershell example. Otherwise the header will not be recognized.


Court72:
@vijayma

#label:"aq-pr-triaged"

Agu1l4r821 (approved):

steved0x:
#sign-off

steved0x (approved):

PR #14078, opened 2024-05-08
Update commit-history.md

SamGrantham:
Adding Clarification that the UI Full History works differently than the Client --Full-History

There have been DevCom and CSS cases regarding the difference and I have confirmed that it is by design.


SamGrantham:
Resubmitting this via the Learn Editror after Reauthing it to my Github account
Please see: https://github.com/MicrosoftDocs/azure-devops-docs-pr/pull/6740

PR #14088, opened 2024-05-30
Remove non working code block

bjompen:
The removed code block defined the `-Trusted` parameter twice and throws an error. The above code block already has the `-Trusted` parameter set, so there is no need to have this example as a separate block.


Court72:
@ramiMSFT

#label:"aq-pr-triaged"

ramiMSFT (approved):

ramiMSFT:
@bjompen Thank you for your contribution! 

#sign-off

PR #14098, opened 2024-07-01
Update conditions.md

aamrun:
The example yaml under job example 2 which is supposed to show the dependency of a following stage on a primary stage is identical to example 1 where the stages are independent. The description of example 2 says it has the condition : succeeded() but this is not present in the yaml. This edit includes that condition in example 2's yaml.


Court72:
@juliakm

#label:"aq-pr-triaged"

juliakm:
Thanks for catching this! 

#sign-off

PR #14112, opened 2024-08-01
Specify -t flag for ssh-keygen

dbeezt:
Azure DevOps requires ssh keys in RSA format, which is becoming deprecated. I just setup a new WSL instance of Ubuntu-24.04 where the default key format generated via ssh-keygen (without flags) is e25519. I think this doc. should be updated to stay current with changing distro behaviour by specifying the type now that RSA isn't guaranteed to be the default.

### DIFF
--- a/docs/repos/git/use-ssh-keys-to-authenticate.md
+++ b/docs/repos/git/use-ssh-keys-to-authenticate.md
@@ -58,7 +58,7 @@ To use key-based authentication, you first need to generate public/private key p
 To generate key files using the RSA algorithm, run the following command from a PowerShell or another shell such as `bash` on your client:
 
 ```powershell
-ssh-keygen
+ssh-keygen -t rsa
 ```
 
 The output from the command should display the following output (where `username` is replaced by your username):


dbeezt:
Hi @Court72, it looks like @vijayma isn't an active reviewer based on this repo.'s history. Would you mind assigning a new reviewer please? Thanks!

Court72:
@bjanousek - As a recent contributor to _docs/repos/git/use-ssh-keys-to-authenticate.md_, can you review the proposed changes?

cc @vijayma

bjanousek (changes requested):
We should instruct users to use one of these variants now:
ssh-keygen -t rsa-sha2-256
ssh-keygen -t rsa-sha2-512

We do not support only the rsa that refers to SSH-RSA (which is SHA1).

I've tried proposing changes via this PR - https://github.com/dbeezt/azure-devops-docs/pull/1

dbeezt:
Merged your change @bjanousek. Much appreciated!

bjanousek (approved):
The change looks good now.

bjanousek:
#sign-off

PR #14124, opened 2024-08-12
Update azure-devops-cli-in-yaml.md

markwragg:
The PowerShell example for "Install Azure CLI and Azure DevOps CLI extension in your pipeline" used Bash instead of PowerShell for its tasks. It is correct in the two code samples above this one, but was incorrect in the block that then combined them together into a single script.

It's a tiny change, the commands are otherwise the same, it just needs to call PowerShell for the script tasks instead of Bash.


chcomley (approved):

chcomley:
#sign-off

PR #14157, opened 2024-10-23
Update migration-prepare-test-run.md

Rajesh-Chenna-rule:
I have noticed a Typo in the section Configure the migration specification file, instead of do it was to


Jak-MS:
@Rajesh-Chenna-rule - thanks for the edit!

PR #14176, opened 2024-11-19
Adding clarification to views note

ryangurn:
This is unclear because the default view can be changed in AzDo but not really at runtime, only on a global level. I am open to any feedback on wording since the original text is causing quite a bit of confusion.


Jak-MS:
@ramiMSFT 

#label:"aq-pr-triaged"

ryangurn:
@Jak-MS, do you have any recommendations on the wording or is that mostly @ramiMSFT who would provide that PR feedback?

Jak-MS:
@ryangurn - @ramiMSFT will respond as time permits to review your suggested changes. Thanks!

ramiMSFT:
@ryangurn Thank you for your contribution. However, this is not accurate. Changing the default view (to "Release" for example) does not enable direct publishing to that view. Packages can only be published to the base feed, where they will be available in the "Local" view.

PR #14205, opened 2025-01-21
Update scheduled-triggers.md

mariselo:
Fix public docs about schedule triggers with always flag. Currently public doc says that the always flag means that schedule builds should be queued only if there are code changes, which is not true. Changes to the pipeline will also cause schedule trigger to be queued.


ttorble:
@steved0x 

#label:"aq-pr-triaged" 

steved0x:
@mariselo Thank you - I will get this updated here as well: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/schedules-cron?view=azure-pipelines

PR #14225, opened 2025-02-05
:memo: correct use of the userID in the API call

verbeek45:
With the given description, mentions weren't created correctly and the user was not linked to the work item. With the ID from the user entitlement the mentioning works.


chcomley (approved):

chcomley:
#sign-off

PR #14253, opened 2025-02-27
Update allow-list-ip-url.md

Ahetejaz:
Microsoft is recommending users to use MDP instead of VMSS hence it would be more beneficial for reader to know that in this page and also provide links to the same.


chcomley (approved):

chcomley:
#sign-off

PR #14290, opened 2025-05-11
Updated doc, removing empty row

SebastianClaesson:
Removing empty entry


v-regandowner:
Looks good; thanks @SebastianClaesson

PR #14316, opened 2025-08-06
Update billing-faq.yml

anmason:
Added warning to the mitigation steps for the Azure resource location being incorrect that the org will revert to free quantities until billing is reconfigured


ttorble:
@anmason did you intend to open this PR in the private repo? It looks like you have access to the private repo, and if so, you should use that for content updates where the content can be staged and validated by the current validation rules. I'll close this PR. Thanks.

PR #14322, opened 2025-08-29
Clarify sprint-257 that stageList can only be used for base YAML and not for templates 

jcheung-dev:
This warning is available at https://learn.microsoft.com/en-us/azure/devops/pipelines/process/runtime-parameters?view=azure-devops#parameter-data-types but not on the release notes.


gloridelmorales:
#sign-off

PR #14351, opened 2026-01-27
Update IP address instructions for GitHub hosted runners

arkoppan:
Updated the link in Line 205 for "here" where it says "using the instructions provided here" from:

https://docs.github.com/en/actions/concepts/runners/github-hosted-runners#ip-addresses

To 

https://docs.github.com/en/actions/reference/runners/github-hosted-runners#ip-addresses


steved0x (approved):

steved0x:
#sign-off

PR #14359, opened 2026-03-03
Use less privileged role to create MDP

igorpupkinable:
`Compute Gallery Image Reader` is less privileged role which is sufficient to create Managed DevOps Pool.
Tested in Switzerland North.


steved0x (approved):

steved0x:
#sign-off

PR #14383, opened 2026-07-02
Fix variable reference for agent IP in key vault access

BalintBanyasz:
The sample currently sets the agentIP variable incorrectly:

```echo "##vso[task.setvariable variable=agentIP]ip"```

This stores the literal string ip instead of the value of $ip. As a result, the later Remove-AzKeyVaultNetworkRule step cannot remove the rule that was added, leaving the agent IP in the Key Vault network rules.

The sample should only use:

```echo "##vso[task.setvariable variable=agentIP]$ip"```

Likewise, the removal step should reference the pipeline variable:

```$ipRange = "$(agentIP)/32"```

This ensures the example works end-to-end.


v-regandowner:
@ramiMSFT 

#label:"aq-pr-triaged"

ramiMSFT (approved):

ramiMSFT:
Thank you for your contribution! 

#sign-off

PR #14397, opened 2026-08-13
Add file transform security note and troubleshooting for blocked xdt:Import

nemanjarogic:
## Summary

File transformations run by the File Transform and Azure App Service / Web App / Function App deployment tasks now process only the **built-in** XDT transform and locator types. The `xdt:Import` element (which loads a custom transform assembly by name or path) and any custom `xdt:Transform` / `xdt:Locator` types are rejected by default, because a transform file that comes from an untrusted package could otherwise load and run arbitrary code on the pipeline agent.

Changes described above are made through: https://github.com/microsoft/azure-pipelines-tasks-common-packages/pull/641

This PR documents that secure-by-default behavior, the recommended alternatives, and the opt-out variable.

## Changes

- **`docs/pipelines/targets/includes/webapp/deploy-options.md`** — added an `[!IMPORTANT]` note beside the File Transform task reference: file transformations allow only built-in XDT types; how to adapt (use built-ins, or pre-apply custom transforms before the package reaches the pipeline); and the `AZP_ALLOW_UNSAFE_XDT_TRANSFORMS` opt-out (flagged as not recommended). This is a shared include, so the note surfaces on the Azure Web App / App Service deployment target pages.
- **`docs/pipelines/troubleshooting/troubleshoot-azure-web-app-deploy.md`** — added a new troubleshooting section, *"XML transformation or variable substitution fails after a file transform security update"* (Symptom / Cause / Resolution), for customers whose custom XDT transforms stop working; updated `ms.date`.

## Why

Secure-by-default: untrusted transform content should not be able to execute code on the agent. The opt-out is documented so pipeline authors with a legitimate need can consciously — and temporarily — restore the previous behavior, with the risk called out.


v-regandowner:
@chcomley 

#label:"aq-pr-triaged"

chcomley (dismissed):

v-dirichards:
@nemanjarogic @chcomley Can you take this PR out of review please? We cannot merge while it's in draft mode. Thanks.

